Issue #16, opened 2018-02-13
GraphQL instrumentation


mxstbr:
GraphQL instrumentation could be super cool.

TL;DR every field and object in an API has it's own _resolver_ (function which _resolves_ the value), so you could instrument those to quickly figure out which resolvers are the slowest. There are existing products that focus exclusively on this (most notably engine.apollographql.com), but having that data in conjunction with all my other tracing data would be :100:

Some references from the @apollographql folks: https://github.com/apollographql/apollo-tracing

southpolesteve:
Just wanted to register my interest for this feature. Following closely!

Lughino:
Same here! Would be a very cool feature!

cyrus-za:
Reigniting this thread. Apollo added DataDog integration. I would love to see Honeycomb on there.

![image](https://user-images.githubusercontent.com/1845861/75148244-2a712800-573a-11ea-96c9-b20c4efc5957.png)

rnrsr:
We love honeycomb, would be great to see this implemented, as some of our apps run a node graphql server.

hboylan:
I've adapted [this HoneycombTracingExtension gist](https://gist.github.com/allyjweir/ffc2cb36ed814d2be6ca78d0db54a1a2) for use in our own Apollo Server. It generally works, but parent spans are missing from some of our deeply nested query fields. Would love to get some feedback from the beeline community!

TylerWanner:
@numso I have just built a working POC based on ur gist.  Thanks for that!  It makes a lot of sense
That said, I feel like there's probably a really clean way to integrate with Apollo Tracing.  I don't understand it as well but I'm very interested

vreynolds:
Hello,

We will be closing this issue as it is a low priority for us. It is unlikely that we'll ever get to it, and so we'd like to set expectations accordingly.

As we enter 2022 Q1, we are trimming our OSS backlog. This is so that we can focus better on areas that are more aligned with the OpenTelemetry-focused direction of telemetry ingest for Honeycomb.

If this issue is important to you, please feel free to ping here and we can discuss/re-open.

PR #24, opened 2018-02-16
first pass at graphql resolver magic

toshok:
graphql's compile step from ES2015 modules results in exports being unconfigurable properties.  since they're unconfigurable, we can't use `shimmer`.  Instead, create a new object and copy everything over, replacing the one entrypoint we need at the moment, the constructor for `GraphQLSchema`.  We walk the type map and wrap all field resolvers.

This technique shamelessly lifted from https://github.com/apollographql/graphql-extensions

Somewhat addresses #16. Adding events for validation and total execution would be nice too.

### DIFF
--- a/lib/magic.js
+++ b/lib/magic.js
@@ -13,6 +13,7 @@ const instrumentedModules = new Set([
   "bluebird",
   "sequelize",
   "child_process",
+  "graphql",
 ]);
 const instrumented = new Map();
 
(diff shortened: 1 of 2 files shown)


Lughino:
Hi all, any news for this PR?
What is missing to get this PR merged?

eanakashima:
@Lughino this was something we were experimenting with, but it's not on our immediate roadmap to finish it up and get it ready to merge. Drop us a line with a bit about your use case at solutions@honeycomb.io if you need it for something?

vreynolds:
Housekeeping: closing this as stale, but will keep the branch as reference.

PR #45, opened 2018-03-24
Tweak README

christineyen:
Goal: contextualize this package (especially compared to a normal SDK).

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 This package instruments your Express/NodeJS application for use with [Honeycomb](https://honeycomb.io). Slice and dice requests by endpoint, status, or even User ID, with zero custom instrumentation required([1](#footnotes)).
 
-Requires Node 8+ and Express for best results. Sign up for a [Honeycomb trial](https://ui.honeycomb.io/signup) to obtain a Write Key before starting.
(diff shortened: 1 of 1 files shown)


toshok:
weird, it appears to have lost my comment? :(

README.md
```diff
@@ -1,12 +1,12 @@
 # Honeycomb NodeJS Magic
 
-An experimental onramp to getting your data into [Honeycomb](https://honeycomb.io) as quickly as possible. With zero custom instrumentation required([1](#footnotes)).
+This package instruments your Express/NodeJS application for use with [Honeycomb](https://honeycomb.io). Slice and dice requests by endpoint, status, or even User ID, with zero custom instrumentation required([1](#footnotes)).
 
-Requires Node 8+.
+Requires Node 8+ and Express for best results. Sign up for a [Honeycomb trial](https://ui.honeycomb.io/signup) to obtain a Write Key before starting.
```
toshok: "for best results" feels like it softens the req to the point where I'm wondering if people will assume it works for node<8 / !express, it just won't work as well.  Am I worrying too much?

christineyen: No, this is legit. I was squinting and handwaving at this part. Should we just say "Requires Node 8+ and Express."?

Node 8+ is a hard requirement; what exactly is necessary to get it working for a non-Express app?

PR #50, opened 2018-04-09
instrument by resolved path, include package versions in events

toshok:
A couple of fixes:

1. cache our instrumented packages by their resolved path, not by their request.  This means that if you have two versions of express in the same app (vanishingly unlikely, but...) you end up with two instrumented versions - but you end up with those two versions of express, not the first one to be instrumented/cached.
2. pull in the package.json files for packages as we instrument them, then pass the `packageVersion` down into the magic.  Include that version in all events the magic sends.  This will let us do nice things like COUNT/P99(durationMs) broken down by express version 🔥

### DIFF
--- a/lib/magic.js
+++ b/lib/magic.js
@@ -51,7 +51,7 @@ const instrumentPreload = (exports.instrumentPreload = () => {
 // arguments are the same as to require.resolve.
 function getPackageVersion(request, options) {
   // treat `require("react-dom/server")` like `require("react-dom")` here
-  if (request.indexOf("/") !== -1) {
+  if (request.includes("/")) {
     request = request.slice(0, request.indexOf("/"));
   }
   try {
@@ -70,7 +70,7 @@ function getPackageVersion(request, options) {
 
 const instrumentLoad = (exports.instrumentLoad = (mod, loadRequest, parent, opts = {}) => {
   if (
-    parent.id.indexOf("node_modules/honeycomb-nodejs-magic/") !== -1 ||
+    parent.id.includes("node_modules/honeycomb-nodejs-magic/") ||
     !instrumentedModules.has(loadRequest)
   ) {
     // no magic here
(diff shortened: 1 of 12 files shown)


eanakashima:
@ChristopherBiscardi you have bandwidth to review this one? I can if not, but I feel like you might have more useful background context.

lib/magic/child_process-magic.test.js
```diff
@@ -4,7 +4,7 @@ const child_process = require("child_process"),
   tracker = require("../async_tracker"),
   event = require("../event");
 
-instrumentChildProcess(child_process);
+instrumentChildProcess(child_process, {});
```
ChristopherBiscardi: would be nice to set this second arg as a default arg instead of passing it in like:

```javascript
let instrumentChildProcess = (child_process, opts = {}) => {
```

but the amount of code in that fn is pretty low, so maybe:

```javascript
  shimmer.wrap(child_process, "execFile", wrapExecLike("execFile", opts.packageVersion || {}));
```

lib/magic/http-magic.test.js
```diff
@@ -4,7 +4,7 @@ const http = require("http"),
   tracker = require("../async_tracker"),
   event = require("../event");
 
-instrumentHttp(http);
+instrumentHttp(http, {});
```
ChristopherBiscardi: same default value comment here. Would be nice to accept `undefined` in the second arg and default to `{}`

lib/magic.js
```diff
@@ -47,6 +48,26 @@ const instrumentPreload = (exports.instrumentPreload = () => {
   });
 });
 
+// arguments are the same as to require.resolve.
+function getPackageVersion(request, options) {
+  // treat `require("react-dom/server")` like `require("react-dom")` here
+  if (request.indexOf("/") !== -1) {
```
ChristopherBiscardi: `includes` is supported back to node 6 (and node 4 is officially unsupported this month).

```
request.includes("/")
```

toshok: sigh, I _never_ remember this method.  #old

PR #51, opened 2018-04-09
schema changes

toshok:
adds `schema.js` which contains all the column names not specified by the magic modules themselves (meta.*, duration, etc.),
and make use of them everywhere.  that way we can bikeshed a bit more easily (without requiring global search and replace.)

While making this change, unify the duration_ms column with express's handling (response_time_ms) and land on durationMs for
everything.

Also, drop all of express's column prefixes, and camelCase its field names instead of snake_case.

### DIFF
--- a/lib/schema.js
+++ b/lib/schema.js
@@ -1,0 +1,10 @@
+/* global exports */
+
+exports.EVENT_TYPE = "meta.type"; // XXX(toshok) rename "type" to "source"?
+exports.REQUEST_ID = "meta.request_id";
+exports.REQUEST_ID_SOURCE = "meta.request_id_source";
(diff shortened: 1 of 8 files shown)


lib/schema.js
```diff
@@ -0,0 +1,10 @@
+/* global exports */
+
+exports.EVENT_TYPE = "meta.type"; // XXX(toshok) rename "type" to "source"?
```
toshok: mostly tagging @maplebed to get his thoughts on these columns.  camelCase for the `meta.` columns too, I'm guessing?

ChristopherBiscardi:
pulling that out was great. 👍

PR #59, opened 2018-07-06
trace propagation

toshok:
marshal trace context (trace_id, parent_id, and custom context) into an http header in both http/https instrumentations, and unmarshal it in the express instrumentation.

The format of the header value is:

```
  ${VERSION};${PAYLOAD}
```

with version 1 of the payload looking like:

```
  trace_id=${incoming trace id},parent_id=${incoming parent span id},context=${base64 encoded stringified JSON object}
```

Both `trace_id` and `parent_id` are required.  `context` can be missing without error, and any other `key=value` is ignored.  If `trace_id` or `parent_id` is missing, or if `context` fails to decode/parse as JSON, the context is dropped.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules/
+yarn-error.log
--- a/lib/event_api/index.js
+++ b/lib/event_api/index.js
@@ -48,6 +48,9 @@ module.exports = {
   addContext(map) {
     return eventAPI.addContext(map);
   },
+  removeContext(key) {
+    return eventAPI.removeContext(key);
+  },
   customContext: {
     add(key, val) {
       eventAPI.addContext({ [schema.customContext(key)]: val });
--- a/lib/event_api/libhoney.js
+++ b/lib/event_api/libhoney.js
@@ -52,7 +52,7 @@ module.exports = class LibhoneyEventAPI {
   }
 
   startRequest(metadataContext, traceId, parentSpanId) {
-    let id = traceId || uuidv4();
+    const id = traceId || uuidv4();
 
     if (this.ds && !this.ds.sample(id)) {
       // don't create the context at all if this request isn't going to send a trace
@@ -64,7 +64,7 @@ module.exports = class LibhoneyEventAPI {
       customContext: {},
       stack: [],
     });
-    return this.startEvent(metadataContext, undefined, parentSpanId);
+    return this.startEvent(metadataContext, id, parentSpanId);
   }
 
   finishRequest(ev) {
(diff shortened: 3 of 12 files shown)


toshok:
Just remembered I need to add some additional logic for handling strings/options/urls in the http(s) instrumentation.

nathanleclaire:
Any chance I could get you to post a quick little code snippet showing how this is meant to be used in code? Will we wrap and instrument common HTTP request modules, or simply provide a way to generate the custom header and leave it to the user to pass along with their request library of choice?

toshok:
> Any chance I could get you to post a quick little code snippet demonstrating how this is meant to be used in code? Will we wrap and instrument common HTTP request modules, or simply provide a way to generate the custom header and leave it to the user to pass along with their request library of choice?

The intention is to support both use cases, although for http usage the automatic case should be all that's required (1 below)

1. The express instrumentation reads from the special `X-Honeycomb-Trace` header by default, and both `http`/`https` instrumentations add the header on outgoing requests.  This happens automatically without any additional programmer work.
2. For non-http transport cases, I've added most of what's needed: `marshalTraceContext`/`unmarshalTraceContext`.  There's likely some bikeshedding that needs to happen here, since I wanted to separate the act of unmarshaling the trace context from installing it at the root of a new trace subtree (which we currently have no API for.)

tredman:
I'll take a look this week!

tredman:
Are there any current examples of what the custom context would be used for?

toshok:
Some snippets here talking about it:

https://github.com/honeycombio/beeline-nodejs/blob/master/docs/CustomContext.md

tredman:
Ok that makes sense now (need to get the other beelines on the same page with how context is passed down to each event).

Regarding passing this along in the trace header: if this is going out with every http/https request, does it expose the user to accidental leaks of sensitive information? Thinking about the user that adds a field for, say, "user email", then makes an unrelated outbound call to some third-party API outside of their control.

toshok:
going to pull some of these changes (like the event_api refactor) out of this PR to make it a little easier to review.

toshok:
alrighty, this should be a little easier to review (basically just has changes that need to be there for propagation to work.)

toshok:
> LGTM. I like the separation of {,un}marshalTraceContext and start{Request,Event}. I don't understand the layout well enough yet - is the idea that {,un}marshal are part of the public API? That seems good, so people can write other instrumentations that can propagate our header format.

yeah, sorry about the lack of front-loading of api plans.  that's been somewhat intentional.  I've been procrastinating on what a public api might look like until all the pieces are in play.  going to have another PR shortly to start the bikeshedding :)

(Oh, and one more re: plugging possible leaks of sensitive information in the trace header payload.)

Re writing other instrumentations that can propagate our header format: it might be a good idea to separate out the header marshaling into a completely separate npm package with just that one .js file (but keep it in this repo) such that it's easier for folks doing manual tracing with just libhoney-js (or their own library) without pulling in the beeline.

PR #65, opened 2018-07-18
event.js is getting a little hard to navigate.  split out the two impls to a subdir

toshok:
sneak in a few other tiny fixes, like ensuring that our http{s} instrumentation work with strings/objects/urls.

### DIFF
--- a/lib/instrumentation/express.test.js
+++ b/lib/instrumentation/express.test.js
@@ -66,7 +66,6 @@ describe("userContext", () => {
             "response.status_code": "200",
           })
         );
-        //        event._apiForTesting().sentEvents.splice(0, 1);
         event._resetForTesting();
         server.close();
         done();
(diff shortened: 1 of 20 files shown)


lib/instrumentation/express.test.js
```diff
@@ -83,14 +64,29 @@ for (let t of tests) {
...
-          });
-          event.apiForTesting().sentEvents.splice(0, 1);
-          done();
-        });
-    });
-  });
-}
+          })
+        );
+        //        event._apiForTesting().sentEvents.splice(0, 1);
```
akvanhar: Should this still be commented out? I'd either remove it or fix the test.

toshok: funky - that was removed at one point.  last night there were so, so many cherry-picks and rebases.  Will fix.

PR #66, opened 2018-07-20
Add a public API

### LINKED ISSUE #12: enable going off-road in the onramp
This onramp is meant to be an easy way to get people started sending events with useful (but generic) data about the node modules they're using.  This obviously only gets people so far.

We need to make it super easy to add additional context to events.  Individual metrics are somewhat nicely served by the `magic.customContext.add(key, value)` api.  But that is only nice for single point in time metrics.  Timers around functions, or worse - async operations, are much more difficult and gross to add.

@nathanleclaire mentions some NR sugar here: https://github.com/honeycombio/honeycomb-nodejs-magic/pull/4#issuecomment-362720033

adding something along those lines would help in the case where a user isn't using a supported top-level module.  Right now the only top-level module is express.  People using less popular frameworks (or heck, the framework they wrote on their own) should be able to get reasonable instrumentation out of it just as easily.

### LINKED ISSUE #40: expose an api that allows users to propagate context if they find problems
Right now we rely on these passthrough instrumentations (e.g. mpromise, bluebird) to propagate our async context around.

One could imagine exposing enough of an api such that #36 was fixable without addition to this repo (I would still want the fix upstreamed, but in the general case users should be able to fix gaps themselves).

toshok:
a lot going on here, but the api is in `lib/api/index.js`.  Could use some more comments there, but the idea is that it should cover a large % of external use-cases, as _well_ as be all an instrumentation author needs.

`Request` is renamed to `Trace`, `Event` is renamed to `Span`.  There are two new functions for those nouns, `with${Noun}`, which does both start/finish wrapped around a synchronous function.  The react-dom-server instrumentation now makes use of `withSpan`.

There's also high resolution timer support `startTimer`/`endTimer`/`withTimer`.

A number of functions from other modules are re-exported here with better names (in particular `bindFunctionToTrace` and `runWithoutTrace`).  These two are both used all over.

The bulk of the change is updating all instrumentations to use the new names (so many "`ev`"s got replaced by `trace` and `span`).

Fixes #12 
Fixes #40


BrianChristie:
I'm trying to get this working with .startSpan and .withSpan, to create additional spans around functions. I'm not sure how the metadataContext works, which startSpan needs. Any chance you could add some how-to docs?

tredman:
This is looking good to me, with my limited exposure to JS. The API looks like everything we'd want and nothing we wouldn't - will be cool to standardize on this in other beelines where possible. 


Would like to see the docs finished before landing. :)

toshok:
if/when we land this, I think labeling it experimental would be a good thing, allowing some bake-in time and to give folks an easier path to trying it out than cloning the repo.

I've got some schema/other changes to make that are breaking enough that a major bump is probably warranted, and we could piggyback off that bump to make the api official.

docs/API.md
```diff
@@ -0,0 +1,381 @@
...
+// childSpan's data has been sent to honeycomb.
+// now parentSpan is the current span
+
+beeline.finishSpan(parentSpan);
+// parentSpan's data has been sent to honeycomb.
+```
+
+#### withSpan()
+
+If you're doing something synchronously (looping, for instance, or using a synchronous node api) you can use `withSpan` to wrap this operation. It safely wraps the invocation of `fn` with `startSpan()` and `finishSpan()` calls. It returns the return value of fn, so can be used in a expression context.
```
christineyen: had a deja vu moment. could we get a line in here that clarifies when you'd use this and when you'd use `withTrace()`?

docs/API.md
```diff
@@ -0,0 +1,381 @@
...
+
+If you're dealing with multiple services (either on the same host or different ones) and want the services to all participate in the same trace, some information about the trace itself needs to be propagated on outbound calls (and then consumed on the other side.) For outbound http/https and inbound express, the propagation happens automatically. The following APIs exist to ease the task of adding propagation to other transports.
+
+#### marshalTraceContext()
+
+```javascript
+beeline.marshalTraceContext();
+```
+
+Returns a serialized form of the current trace context (including the trace id and the current span), encoded as a string. The format is documented at https://github.com/honeycombio/beeline-nodejs/blob/master/lib/propagation.js#L16
```
christineyen: did you know github lets you link to a "permalink"? :D let's do that in case line numbers change at some point in the future: https://github.com/honeycombio/beeline-nodejs/blob/9a6bb4203869546914d5d623913e3f810f81eed2/lib/propagation.js#L16

docs/API.md
```diff
@@ -0,0 +1,381 @@
...
+
+```javascript
+beeline.startTrace(metadataContext[, withTraceId, withParentSpanId])
+```
+
+Starts a new local trace and initializes the async context propagation machinery. Most other API calls require a trace to be active to actually do anything (i.e. calling `startSpan` does nothing if there's no trace.) Returns a reference to the trace, and installs the root span as the current span.
+
+`metadataContext` is a map of initial properties of the root span.
+
+For an explanation of `withTraceId` and `withParentSpanId`, see "Interprocess trace propagation" below.
```
christineyen: s/Interprocess trace propagation/(Interprocess trace propagation)[#interprocess-trace-propagation]/

(i think this will work)

docs/API.md
```diff
@@ -0,0 +1,381 @@
...
+
+  try {
+    await handleSomething(restOfPayload);
+  } finally {
+    beeline.finishTrace();
+  }
+});
+```
+
+####
```
christineyen: (cleanup?)

docs/API.md
```diff
@@ -0,0 +1,381 @@
...
+
+There are two axes to useful traces: depth and width. Depth is a measure of how many operations are performed, and width is a measure of how much information there is to associate with each operation. Depth is addressed by the startSpan/finishSpan. Width is addressed by the methods in this section.
+
+#### addContext()
+
+```javascript
+beeline.addContext(contextMap);
+```
+
+Adds all key/value pairs in `contextMap` as fields on the current span. Used primarily from beeline instrumentation (and use a cross language beeline namespace.)
```
christineyen: dumb question: do fields typically propagate down to children by default? no, right? (could we add a note to that effect here?)

docs/API.md
```diff
@@ -0,0 +1,381 @@
...
+beeline.removeContext(fieldName);
+```
+
+#### customContext.add()
+
+```javascript
+beeline.customContext.add(key, value);
+```
+
+Adds a single key/value pair as a field on the current span. The key is automatically prefixed with `app.` to set it apart from (and keep from conflicting with) the instrumentation-added fields.
```
christineyen: Does this prefix the column name, while `addContext()` doesn't? Can we add something to the `addContext()` description, after the "Used primarily from beeline instrumentation..." sentence, to the effect of "Adding fields to the context will add the provided keys to the top-level payload for a span. We recommend using `customContext.add()` for metadata specific to your application."?

docs/API.md
```diff
@@ -0,0 +1,381 @@
...
+example:
+
+````javascript
+myDBLibrary.query("select * from table",
+                  beeline.bindFunctionToTrace(rows => {
+                    // inside this function the trace is guaranteed to be
+                    // the same as the the active trace when myDBLibrary.query
+                    // was called, even if some pattern in myDBLibrary causes
+                    // the context to be lost.
+                  }));
```
christineyen: missing '```'?

docs/API.md
```diff
@@ -0,0 +1,381 @@
...
+beeline.customContext.remove("userName");
+```
+
+#### schema
+
+_[TODO more here, but it should be 99% of use to instrumentation authors, not beeline users]_
+
+### Async Context Bookkeeping
+
+The beeline uses nodejs's builtin `async_hooks` module to ensure its trace context is propagated through async calls, but there are some common patterns that break this magic (The most common is a worker/connection pool abstraction seen in many db packages).
```
christineyen: These examples are awesome - and a lazy way to make them more awesome with minimal work would be linking to specific instrumentations (maybe in https://github.com/honeycombio/beeline-nodejs/tree/toshok.public-api/lib/instrumentation?) that exhibit or take advantage of certain parts of this API, if applicable.

toshok: good call.  will do that.

docs/API.md
toshok: 👍

docs/API.md
toshok: correct, they don't.  _These_ fields don't, but custom context does.  Need to make a note of that as well.  We need to figure out the context propagation (intra-process, inter-process) story.

docs/API.md
toshok: 👍

docs/API.md
```diff
@@ -0,0 +1,381 @@
...
+
+example:
+
+```javascript
+let trace = beeline.startTrace({
+  task: "writing a file",
+  filePath,
+});
+fs.writeFile(filePath, fileContents, err => {
+  beeline.finishTrace(trace);
```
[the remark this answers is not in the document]
toshok: 👍

docs/API.md
```diff
@@ -0,0 +1,381 @@
...
+);
+
+let fib = beeline.withTrace(
+  {
+    task: "computing fibonacci number",
+    n,
+  },
+  () => computeFib(n)
+);
+console.log(`The answer is ${fib}.`);
```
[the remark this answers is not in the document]
toshok: 👍

docs/API.md
```diff
@@ -0,0 +1,381 @@
...
+  task: "writing a file",
+  filePath,
+});
+fs.writeFile(filePath, fileContents, err => {
+  beeline.finishSpan(trace);
+});
+
+let span1 = beeline.startSpan({
+  name: "parent span",
+});
```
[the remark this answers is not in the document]
toshok: oh, I think this was going to be another example.  removing.

docs/API.md
```diff
@@ -0,0 +1,381 @@
...
+The beeline assumes that within a process, all spans are fully unclosed by their parent span. If you finish a parent span before all children have been finished we'll emit a warning.
+
+example:
+
+```
+let parentSpan = beeline.startSpan({
+  name: "parent span",
+});
+
+let childSpan = beeline.startSpan({
```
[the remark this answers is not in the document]
toshok: added a lot of comments here, hopefully addresses blindingly obvious :)

docs/API.md
toshok: 👍

docs/API.md
```diff
@@ -0,0 +1,391 @@
...
+    parentSpanId
+  );
+
+  try {
+    await handleSomething(restOfPayload);
+  } finally {
+    beeline.finishTrace();
+  }
+});
+```
```
samstokes: This example could be read as suggesting that it's `service2`'s job to call `finishTrace` for the _entire_ distributed trace, i.e. that `service1` no longer needs to call `finishTrace` itself. I assume that's not true, and that each service needs to call `startTrace` and `finishTrace` unless they're in an express handler that does it for them?

If so it might be clearer to just include the `startTrace` and `finishTrace` calls in the `service1` example, instead of saying "assuming we're in a trace", and then maybe add a comment "you can skip these if you're in an express handler".

samstokes:
The API looks great, and similar enough to the APIs in Python and (experimentally) Ruby as to not confuse people who are using multiple Beelines. I agree `startSpan` makes more sense than `newEvent` or something. (Although I wonder if we also want some facility to send a fire-and-forget event/span that can disable the warning if the child outlives its parent span?)

Now to actually read the code... :)

docs/API.md
toshok: yep, your assumption is correct.  I'd used the words "local trace" somewhere in the doc but I'm not sure how well that communicates things (especially lacking any text re: how local traces are stitched together into a larger distributed trace).

will futz with example 👍

lib/index.js
```diff
@@ -1,19 +1,21 @@
...
 
   return configure;
 }
 
-configure.asyncTracker = require("./async_tracker");
-configure.customContext = event.customContext;
-
-configure.marshalTraceContext = propagation.marshalTraceContext;
-configure.unmarshalTraceContext = propagation.unmarshalTraceContext;
+// copy the api as properties on the configure function
```
samstokes: Hmm, this made me double-take when reading the code. So the intended usage would look something like

```javascript
const beeline = require('honeycomb-beeline')({
  writeKey: 'abc',
  // ...
});

beeline.startTrace();
```

I guess that makes sense at point of use. Is it your sense that Node folks are used to APIs that look like that - invoking the result of a `require` and then calling additional methods on the result of that invocation? I guess it just looks like a constructor invocation.

lib/index.test.js
```diff
@@ -1,16 +1,14 @@
 /* eslint-env node, jest */
 test("the default export from the package has asyncTracker and customContext properties", () => {
```
samstokes: Maybe change the test description now we're no longer testing for those properties?

lib/index.js
toshok: yeah, there are a few packages that do that sort of thing, including `debug`.  It makes sense here I think because of the global nature of the async hooks

lib/index.test.js
toshok: 👍

docs/API.md
```diff
@@ -0,0 +1,391 @@
...
+const beeline = require("honeycomb-beeline")();
+
+/* beeline.traceActive() works, as do all the other calls below */
+```
+
+There are a few surfaces to the API:
+
+1.  Traces and spans
+2.  Interprocess trace propagation
+3.  Adding context to spans, including timers
```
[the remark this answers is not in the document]
toshok: 🤦‍♂️ on it.  Going to remove ", including timers" from here, and add some rudimentary docs for the calls

docs/API.md
```diff
@@ -0,0 +1,391 @@
+# Public API
+
+The beeline does a lot of magic for you, but there will probably be parts of your application where you'll want to add
+some additional instrumentation. There might also be problems with async context propagation that you'll need to work
+around while waiting for a bug fix. We now have a public API for you to use that addresses both those use-cases. It'll
+also help if you want write a custom instrumentation for your favorite npm package :)
```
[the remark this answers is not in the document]
toshok: 👍

lib/api/index.js
```diff
@@ -0,0 +1,137 @@
...
+  // a couple of useful functions for either forcing a function to run within a trace, or
+  // force it to run outside of it.  Mostly useful for writing instrumentation.
+  bindFunctionToTrace(fn) {
+    return tracker.bindFunction(fn);
+  },
+  runWithoutTrace(fn) {
+    return tracker.runWithoutTracking(fn);
+  },
+
+  // these are exposed just for tests or instrumentation.  don't use them from outside.
```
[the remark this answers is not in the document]
toshok: yep, although there's no mechanism for adding instrumentations outside this package (yet).

still, good suggestion. 👍

PR #81, opened 2018-08-07
Fix bugs url in askForIssue

toshok:
in our package.json we used to have:
```
  "bugs": "https://...."
```

and the code as written in `lib/api/libhoney.js` assumed that structure.  It worked fine locally (and when npm installing using a file path since that symlinks to the beeline source dir).  But once things are published to npm and installed from there, the `bugs` shorthand is expanded out:

```
  "bugs": {
    "url": "https://...."
  }
```

and askForIssue then prints out, e.g.:

```
honeycomb-beeline error: finishing an event with unfinished nested events. almost certainly not what we want.
please paste this message (everything between the "----" lines) into an issue
at [object Object].  feel free to edit
.
.
.
```

don't use the shorthand in our package.json at all, and modify libhoney.js to no longer assume it.

### DIFF
--- a/lib/api/libhoney.js
+++ b/lib/api/libhoney.js
@@ -197,7 +197,7 @@ module.exports = class LibhoneyEventAPI {
     logger(`-------------------
     ${pkg.name} error: ${msg}
     please paste this message (everything between the "----" lines) into an issue
-    at ${pkg.bugs}.  feel free to edit
+    at ${pkg.bugs.url}.  feel free to edit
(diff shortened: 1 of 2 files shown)


toshok:
unsure where it happens.  I'd assume it happens on publish on npm's side but that's just a guess.

Thankfully this is the only thing we pull from package.json (other than name/version, which are pretty clearly just strings), so again not sure if there are others :)

PR #82, opened 2018-08-08
Remove more magic

garthk:
Strikes me `instrumentPreload` is doubling up on heavy lifting `asyncHook.enable()` in `async_tracker.js` already takes care of:

> Installing async hooks via `async_hooks.createHook` enables promise execution tracking

 – [Node 8 `async_hooks` documentation](https://nodejs.org/docs/latest-v8.x/api/async_hooks.html#async_hooks_promise_execution_tracking)

Changes:

* Enable ES2017 linting
* Add `async / await` test
* Add manual Promise chain test
* Remove `instrumentPreload`

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,7 +4,7 @@
   },
   "extends": ["eslint:recommended"],
   "parserOptions": {
-    "ecmaVersion": 2015
+    "ecmaVersion": 2017
   },
   "rules": {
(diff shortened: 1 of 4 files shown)


toshok:
niiiice.  I could have sworn I tested native promises and they didn't work.  Wonder if it was a later 8.x release then the one I was using that fixed them.  Either way 🎉

PR #83, opened 2018-08-08
Add .editorconfig

garthk:
**I nearly committed code with trailing whitespace.** If your facial expression right now resembles the poor fella in Edvard Munch's *The Scream*, well, aren't you glad there are [ways to make sure it never happens again][phew].

Speaking of: He's not screaming. [Nature is][really].

[phew]: https://editorconfig.org/#download
[really]: https://en.wikipedia.org/wiki/The_Scream#Sources_of_inspiration

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+root = true
+
+[*]
+charset = utf-8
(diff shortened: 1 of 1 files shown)


toshok:
:TIL: about `.editorconfig`

now I need to get vscode using it.

Issue #84, opened 2018-08-09
Dirty working directory after npm install


garthk:
[`prettier` is pretty opinionated][so].

[So is `npm`][much].

We've set them up to fight each other forever.

`lint-staged` forces `prettier` to change `package.json` even if it's listed in `.prettierignore`, so all commits will have long `package-lock.json` lines wrapped and short `package.json` lists collapsed into a single line.

When we check out and `npm install`, `npm` reverses the change. And, so, _ad infinitum_.

[so]: https://github.com/npm/npm/issues/17973#issuecomment-318559060
[much]: https://prettier.io/docs/en/option-philosophy.html

vreynolds:
Closing this issue as there has been no actively for a while. If you feel this should be re-visited, please reopen.

PR #85, opened 2018-08-09
Avoid dirty package*json after npm install (#84)

garthk:
Prior to this patch, `prettier` and `npm` were fighting; see #84 for details.

Fixes:

* Unwrap lines in `package-lock.json`
* Uncollapse lists in  `package.json`
* Set `lint-staged` options to ignore `package.json` and `package-lock.json`

Left alone:

* Version specs in `package-lock.json`, which `npm` 5.6.0 feels the need to update

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,8 +7,7 @@
     "@babel/code-frame": {
       "version": "7.0.0-beta.46",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.46.tgz",
-      "integrity":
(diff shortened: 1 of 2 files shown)


garthk:
I'd have tweaked the options so you could use `prettier` to enforce other options `npm` might leave alone, but couldn't find any such options, nor any options to stop the fight at their end. They own their attitude problem, and good on them, but our only way to stop the fight is to disable them for these two files.

garthk:
Bah! Spurious build failure.
```
OCI runtime exec failed: exec failed: container_linux.go:348: starting container process caused "process_linux.go:86: executing setns process caused \"exit status 21\"": unknown
```

MikeGoldsmith:

garthk:
Fair call.

Issue #86, opened 2018-08-10
conditional/version-specific instrumentation


toshok:
Node 10.0.0 - 10.4.0 breaks async hooks wrt `Promise` usage.

We need the extra promise instrumentation only in that case.  Doing a simple version check should be sufficient (or do we care since it's a really simple case?)

I'd also really rather not litter up instrumentation files with millions of version checks (both for nodejs and for the packages being instrumented.)

should we have minimum-version file names?  ex:

```
lib/instrumentation/express-4.16.3.js
lib/instrumentation/express-4.0.0.js
lib/instrumentation/express.js
```

where we pick all that match and apply them (in some order)?  ugh.

JamieDanielson:
We’re going to close this for now as there has been no recent activity. Feel free to request a re-open if it’s still relevant.

Issue #71, opened 2018-07-24
native node async/await test


toshok:
context doesn't seem to get propagated for native node async/await usage.

toshok:
This looks relevant https://github.com/nodejs/node/issues/20404

BrianChristie:
Looks like nodejs/node#20404 is a duplicate of nodejs/node#20274

[One of the comments on 20274](https://github.com/nodejs/node/issues/20274#issuecomment-395200647) says:
> It looks like this is fixed in 10.4.0 with the update of V8 from 6.6 to 6.7

vreynolds:
Hello,

We will be closing this issue as it is a low priority for us. It is unlikely that we'll ever get to it, and so we'd like to set expectations accordingly.

As we enter 2022 Q1, we are trimming our OSS backlog. This is so that we can focus better on areas that are more aligned with the OpenTelemetry-focused direction of telemetry ingest for Honeycomb.

If this issue is important to you, please feel free to ping here and we can discuss/re-open.

PR #87, opened 2018-08-10
fix async hooks breakage on early v10

toshok:
Turns out the code in #82 is necessary for at least early v10 usage (see https://github.com/honeycombio/beeline-nodejs/issues/71#issuecomment-411852263).

We do need a better story about conditional/version-specific instrumentation?  see #86.

this PR reverts #82 and adds node 10.0.0 to the build matrix (which fails without this revert)


toshok:
cc @garthk

PR #88, opened 2018-08-27
Update README.md

dgolant:
Noticed the "Automatically instrumented packages" anchor link in the README wasn't working,
this should fix it.


samstokes:
Good catch, thank you!

PR #91, opened 2018-10-20
run `npm audit fix --force`

toshok:
Update a lot of our deps (most importantly the transitive `randomatic` and `lodash` that were causing scary security warnings on the repo.)

before:

image

after:

image


toshok:
self-accepting this :)

PR #94, opened 2018-10-24
Support async/concurrent spans

### LINKED ISSUE #93: Trace concurrent actions
Beeline assumes that, within a trace, all actions are sequential, and keeps an internal stack of spans to generate parent-child relationships. This assumption breaks down with e.g. concurrent database queries. Is it possible to trace those events and preserve the correct span containment?

toshok:
Fixes #93 

Add `startAsyncSpan` to the API and only to the libhoney implementation for the time being (i.e. no tests).  It doesn't work the same as `startSpan`, unfortunately.  This is due to my understanding of how async context gets propagated by `async-hooks` - I may be wrong, so I'll keep digging there.

the signature is `startAsyncSpan(metadataContext, spanFn)`.  `metadataContext` is the same as with `startSpan`, and `spanFn` looks similar to `withSpan` but it's not.  it's called with a single argument - the asyncSpan itself.

From my motivating example in #93, here's `slowSpan` when using normal spans:

```
function slowSpan() {
  let asyncSpan = beeline.startSpan({ name: "slow" }); // should be startAsyncSpan here
  return new Promise(resolve => {
    setTimeout(() => {
      beeline.withSpan({ name: "slow-child" }, () => {});
      beeline.finishSpan(asyncSpan);
      resolve();
    }, 5000);
  });
}
```

and here's the converted form:

```
function slowSpan() {
  return beeline.startAsyncSpan(
    { name: "slow" },
    asyncSpan =>
      new Promise(resolve => {
        console.log("in slow promise");
        setTimeout(() => {
          beeline.withSpan({ name: "slow-child" }, () => {});
          beeline.finishSpan(asyncSpan);
          resolve();
        }, 5000);
      })
  );
}
```

The reason for passing the `asyncSpan` into this function is that the `finishSpan` call _must_ happen inside the async context created and installed around this inner function.  we create a new tracking context in `startAsyncSpan`, give it the same trace id, and install the async span as the root span in the new context's stack.  doing it this way means that other non-async child spans will be pushed onto the correct stacks and get the correct parent span id.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -123,7 +123,32 @@ let span = beeline.startSpan({
 });
 fs.writeFile(filePath, fileContents, err => {
   beeline.customContext.add("fileError", err.toString());
-  beeline.finishSpan(trace);
+  beeline.finishSpan(span);
+});
+```
+
+```javascript
+beeline.startAsyncSpan(metadataContext, spanFn);
+```
+
+Starts a new async span within an existing trace. This new span is added as a child of the current span, and recorded as the current
+span within the context of calling `spanFn`. Outside of `spanFn` the current span is unchanged. The newly created span is passed as
+the only argument to `spanFn` (to be used in `finishSpan` below.)
+
+Async spans are the norm within the beeline-packaged instrumentation, and you should use them for all asynchronous operations.
+
+example:
+
+```javascript
+beeline.startAsyncSpan({
+    task: "writing a file",
+    filePath,
+}, span => {
+       fs.writeFile(filePath, fileContents, err => {
+           beeline.customContext.add("fileError", err.toString());
+           beeline.finishSpan(span);
+       });
+   }
 });
 ```
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -127,6 +127,10 @@ fs.writeFile(filePath, fileContents, err => {
 });
 ```
 
+#### startAsyncSpan
+
+_(unstable API - possibly changing in the future)_
+
 ```javascript
 beeline.startAsyncSpan(metadataContext, spanFn);
 ```
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -75,6 +75,10 @@ module.exports = {
     }
   },
 
+  startAsyncSpan(metadataContext, spanFn) {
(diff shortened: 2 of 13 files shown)


toshok:
the complete working motivating example from #93 is below:

```
/* eslint-env node */
let beeline = require("honeycomb-beeline")();

// this doesn't currently work.  when it does, the trace should look like:
//
//    + - root
//    |
//    + - + - fast
//    |   |
//    |   + - fast-child
//    |
//    + - + - slow
//        |
//        + - slow-child

function slowSpan() {
  return beeline.startAsyncSpan(
    { name: "slow" },
    asyncSpan =>
      new Promise(resolve => {
        console.log("in slow promise");
        setTimeout(() => {
          beeline.withSpan({ name: "slow-child" }, () => {});
          beeline.finishSpan(asyncSpan);
          resolve();
        }, 5000);
      })
  );
}

function fastSpan() {
  return beeline.startAsyncSpan(
    { name: "fast" },
    asyncSpan =>
      new Promise(resolve => {
        console.log("in fast promise");
        setTimeout(() => {
          beeline.withSpan({ name: "fast-child" }, () => {});
          beeline.finishSpan(asyncSpan);
          resolve();
        }, 1000);
      })
  );
}

let rootSpan = beeline.startTrace({ name: "root" });

// the order of fast/slow is important here.  we want the fast span to be first
// on the context stack such that it'll finish before the slow one and we'll get
// the error.
Promise.all([fastSpan(), slowSpan()]).then(() => beeline.finishTrace(rootSpan));
```

this PR includes a console.log that prints out the JSON event payload, and can be used to verify that spans will be stitched together into the correct tree form.

toshok:
One further comment from offline was that it'd be nice if the api auto finished the span.  It can't do this in general, but if we make `beeline.finishSpan` not emit an error if a span is double-finished, we _can_ do it for promise return values.  like:

```
let dbResult = await beeline.startAsyncSpan({ ... }, () => database.apiReturningAPromise());
```

Not going to include it in this PR because double-finishing spans might be something we want to flag as an error (this is something we currently do, since the second finisher won't find the event payload in the context stack.)

PR #97, opened 2018-11-07
fix(Request): Passes sampleRate with events

### LINKED ISSUE #95: sample rate not being passed with events
At the moment, we aren't sending a sampleRate in any events, even if the deterministic sampler is in use.

Should check how the go beeline is handling this and duplicate behavior.

panitchkov:
Closes #95


toshok:
Hm, this doesn't set the sample rate on the traces based on whether or not the beeline sampled them - it globally samples events sent from the beeline based on an environment variable.  I don't think this is a useful way to configure sampling.

PR #98, opened 2018-11-07
include deterministic sampler sampleRate with events

### LINKED ISSUE #95: sample rate not being passed with events

toshok:
If we have a deterministic sampler configured, include its sample rate with events that we send.  also switch to using `sendPresampled` for events, instead of `send`.

fixes #95


toshok:
added a test for sampleRate = 1 when there _isn't_ a deterministic sampler, but not quite sure how to add one for the deterministic sampler case.  loop until `startTrace` returns non-null, then send that and verify sample rate?

toshok:
added the aforementioned test.  turns out it was pretty easy, and works :)

PR #105, opened 2018-12-27
Provide span to withSpan()'s fn param

christineyen:
Two compounding incentives:
1. I'm trying to set a custom field on a span **within** a trace, and (AFAICT) `beeline.customContext.add` adds the key/value pair to all active spans within that trace - not just the currently active span.
2. I'm wrapping a simple `return x` function body and want to use `withSpan` instead of `startSpan`.

Is it correct that the only way to set a field on a span **within** a trace is to explicitly modify the `span` map? e.g. `span['app.my_custom_field'] = "my_custom_value"`? The docs don't have an example and I can't quite figure out the right incantations to apply the field to the child span only.

... so this fix (providing the `span` to the body of the `withSpan` `fn`) seems like a nicety that'll help with that, though not totally address incentive number 1.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -196,7 +196,7 @@ beeline.finishSpan(parentSpan);
 
 #### withSpan()
 
-If you're doing something synchronously (looping, for instance, or using a synchronous node api) you can use `withSpan` to wrap this operation. It safely wraps the invocation of `fn` with `startSpan()` and `finishSpan()` calls. It returns the return value of fn, so can be used in a expression context.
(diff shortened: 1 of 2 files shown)


toshok:
`beeline.addContext` should do what you want (from the docs):

--- 

```javascript
beeline.addContext(contextMap);
```

Adds all key/value pairs in `contextMap` as toplevel fields on the current span. Should only be used when writing your own instrumentation. Context added with this method is _only_ attached to the current span. We recommend using `customContext.add()` for application-specific context, as `customContext.add()`-added context will be attached to all spans sent after it was added.

---

we could probably make that sound a little less scary.

christineyen:
Closing, will submit a separate PR for some unrelated docs changes (and a related PR on our docs repository to update our honeycomb.io docs with this info)

PR #107, opened 2019-01-06
fix mongodb instrumentation when using Promises

### LINKED ISSUE #89: mongodb instrumentation promises doesn't seem to propagate results / errors
There appears to be an issue with the mongodb instrumentation where promises don't return any value. This means any promise payloads & errors get swallowed

geekflyer:
The current mongodb instrumentation breaks applications which use any collection method which returns a Promise.

This MR fixes that by:
1. properly `return`ing the promise in all places
2. return the original promise (rv) and add instrumentation on a seperate promise chain
3. Replace `.then(success,fail)` anti-pattern with `.then(success).catch(fail)`. see http://bluebirdjs.com/docs/anti-patterns.html#the-.then for an explanation why this is an anti-pattern.

Besides a few other things this MR also fixes #89

### DIFF
--- a/lib/instrumentation/mongodb.js
+++ b/lib/instrumentation/mongodb.js
@@ -41,7 +41,7 @@ function instrumentMongodb(mongodb, opts = {}) {
 
         let eventName = `collection.${name}`;
 
-        api.startAsyncSpan(
+        return api.startAsyncSpan(
           {
             [schema.EVENT_TYPE]: "mongodb",
             [schema.PACKAGE_VERSION]: opts.packageVersion,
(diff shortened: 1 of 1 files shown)


toshok:
💯  to everything except #3:

it's there for compat.  At one point while building the beeline, a version of something (my failing memory suggests mongoose + mpromise, but it's been a while) configured the mongo db driver to use a promise implementation that didn't implement `.catch` at all.

geekflyer:
@toshok adressed ur comment.

toshok:
🎉

Issue #104, opened 2018-12-23
beeline's http instrumentation breaks @sentry/node and potentially other modules with `Error: write EPROTO 140735813014400:error:140770FC:SSL`


geekflyer:
Honeycomb beeline seems to break @sentry/node's error reporting.
When using them together, every error seems to get reported as:
```
Error: write EPROTO 140735813014400:error:140770FC:SSL routines:SSL23_GET_SERVER_HELLO:unknown protocol:../deps/openssl/openssl/ssl/s23_clnt.c:827:
```

## Reproduction:
- node.js version: v8.12.0
- `npm start` and then run `curl http://localhost:3000`.

```json
{
  "main": "app.js",
  "scripts": {
    "start": "node app.js"
  },
  "dependencies": {
    "@sentry/node": "4.4.1",
    "bluebird": "3.5.3",
    "express": "4.16.4",
    "honeycomb-beeline": "1.2.1"
  }
}
```

```js
require('honeycomb-beeline')({
  writeKey: 'REDACTED',
  dataset: 'api-traffic',
  serviceName: 'error-tracking'
});
global.Promise = require('bluebird');
Promise.config({
  longStackTraces: true
});
const Sentry = require('@sentry/node');
Sentry.init({ dsn: 'https://REDACTED@sentry.io/REDACTED' });

const app = require('express')();

app.use(Sentry.Handlers.requestHandler());

app.use((req, res, next) => {
  throw new Error('foo');
});

app.use(Sentry.Handlers.errorHandler());

app.listen(3000);
```

### expected:
`Error('foo')` reported to sentry

### actual:
```
Error: write EPROTO 140735813014400:error:140770FC:SSL routines:SSL23_GET_SERVER_HELLO:unknown protocol:../deps/openssl/openssl/ssl/s23_clnt.c:827:

  File "net.js", line 868, col 14, in WriteWrap.afterWrite [as oncomplete]
  File "/Users/christian/dev/solvvy-apis/error-tracking/node_modules/@sentry/node/dist/transports/base.js", line 40, col 39, in HTTPSTransport.<anonymous>
    return [2 /*return*/, new Promise(function (resolve, reject) {
  File "/Users/christian/dev/solvvy-apis/error-tracking/node_modules/tslib/tslib.js", line 133, col 27, in step
    op = body.call(thisArg, _);
[...]

PR #109, opened 2019-01-07
[https] bind wrapped_cb so we get the proper context in node8

### LINKED ISSUE #106: Error finishing spans in https requests
**Node version: 8.12.0**

**Code:**
```
const beeline = require('honeycomb-beeline')({
  writeKey: 'WRITE_KEY',
  dataset: 'DATASET',
});

require('http');
require('https');
const request = require('request-promise-native');

const trace = beeline.startTrace({
  task: 'http request',
});

request({
  method: 'GET',
  url: 'https://www.google.ro',
}).then(() => {
  beeline.finishTrace(trace);
});
```

**Error:**
```
honeycomb-beeline:event -------------------
  honeycomb-beeline:event     honeycomb-beeline error: no payload for event we're trying to finish (stack is empty).
  honeycomb-beeline:event     please paste this message (everything between the "----" lines) into an issue
  honeycomb-beeline:event     at https://github.com/honeycombio/beeline-nodejs/issues.  feel free to edit
  honeycomb-beeline:event     out any application stack frames if you'd rather not share those
  honeycomb-beeline:event     Error
  honeycomb-beeline:event     at LibhoneyEventAPI.askForIssue (.../node_modules/honeycomb-beeline/lib/api/libhoney.js:249:7)
  honeycomb-beeline:event     at LibhoneyEventAPI.finishSpan (.../node_modules/honeycomb-beeline/lib/api/libhoney.js:149:12)
  honeycomb-beeline:event     at Object.finishSpan (.../node_modules/honeycomb-beeline/lib/api/index.js:67:20)
  honeycomb-beeline:event     at ClientRequest.wrapped_cb (.../node_modules/honeycomb-beeline/lib/instrumentation/https.js:49:17)
  honeycomb-beeline:event     at emitOne (events.js:121:20)
  honeycomb-beeline:event     at ClientRequest.emit (events.js:211:7)
  honeycomb-beeline:event     at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:544:21)
  honeycomb-beeline:event     at HTTPParser.parserOnHeadersComplete (_http_common.js:117:17)
  honeycomb-beeline:event     at TLSSocket.socketOnData (_http_client.js:440:20)
  honeycomb-beeline:event     at emitOne (events.js:116:13)
  honeycomb-beeline:event     current request context:
  honeycomb-beeline:event     ------------------- +0ms
```

toshok:
Fixes #106 

On node v8, the http module calls through to the http module, which means the instrumented https calls through to the instrumented http.  In the past this wasn't a problem.  The user would end up with 2 spans for each https call but it would still work.

async spans change things slightly.  The reason is that they each get their own tracking context.  Tracking contexts are communicated back up the callback chain, which means the https wrapped_cb is executed using the http context.  Again, this was fine pre-async span since all contexts were the same (there was only one.)  It breaks now, though, since we want https's callback to be running in its own context.

Fix is simple: Drop a `tracker.bindFunction` in there to ensure it runs in the proper context.

We don't need to do this for all async spans because afaik this is the only async span that calls through to another async span.

Added a test which hopefully only runs on node 8 (using a semver check.)  Not sure how I feel about this practice, but given behavior differences I guess it's the only way? /shrug

### DIFF
--- a/lib/instrumentation/https.js
+++ b/lib/instrumentation/https.js
@@ -2,7 +2,8 @@
 const url = require("url"),
   shimmer = require("shimmer"),
   api = require("../api"),
-  schema = require("../schema");
+  schema = require("../schema"),
+  tracker = require("../async_tracker");
 
 let instrumentHTTPS = (https, opts = {}) => {
   shimmer.wrap(https, "get", function(_original) {
@@ -45,12 +46,18 @@ let instrumentHTTPS = (https, opts = {}) => {
           url: url.format(options),
         },
         span => {
-          let wrapped_cb = function(res) {
+          // in node 8.x, https calls into http.  currently we don't have a way
+          // to communicate between instrumentations (ideally https would somehow
+          // tell http that it shouldn't create the span), so this creates an additional
+          // async span for the http call. the problem is that our wrapped_cb will be
+          // called with the http context active, not ours.  so we have to bind our
+          // callback here.
+          let wrapped_cb = tracker.bindFunction(function(res) {
             api.finishSpan(span, "request");
             if (cb) {
               return cb.apply(this, [res]);
             }
-          };
+          });
 
           // shallow clone options.headers, adding our tracing headers in
           options.headers = Object.assign({}, options.headers, {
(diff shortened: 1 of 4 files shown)


christineyen:
does this also address https://github.com/honeycombio/beeline-nodejs/issues/104? 🤞

toshok:
I was hoping, but hadn't checked yet

toshok:
unfortunately it doesn't.  going to look at that today.

PR #116, opened 2019-01-23
custom context should be sent to async spans as well

toshok:
Async spans are implemented essentially as separate traces with the async span as the root span (though they have the same trace id as their parent, and the parent_id is set correctly on this root span.)

We were initializing the `customContext` map the same way the root trace does, though, which means it was a new empty object, `{}`.  This means `customContext` isn't shared correctly, and the async spans won't have those fields added.

Instead of initializing it to a new object, just share (by ref) the same object across all async spans and their parent trace.

### DIFF
--- a/lib/api/libhoney.js
+++ b/lib/api/libhoney.js
@@ -112,13 +112,13 @@ module.exports = class LibhoneyEventAPI {
   startAsyncSpan(metadataContext, spanFn) {
     let parentId;
     let spanId = uuidv4();
-    let context = tracker.getTracked();
-    if (context.stack.length > 0) {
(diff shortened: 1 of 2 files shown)


lib/api/libhoney.js
```diff
@@ -128,8 +128,8 @@ module.exports = class LibhoneyEventAPI {
     }
 
     let newContext = {
-      id: context.id,
-      customContext: {},
+      id: parentContext.id,
+      customContext: parentContext.customContext,
```
toshok: the rest of the change is renaming - this is the only substantive change, sharing the object byref instead of creating a new empty object.

PR #125, opened 2019-03-02
Add presend and sampler hooks implementation.

allyjweir:
### Aim
To allow users of the library to define their own `samplerHook` function that takes an event's data and can make a sampling decision based on that. Also allow a `presendHook` function to be defined and called to enrich an event with further context.

### Reasoning (for samplerHook)
Dynamic sampling is something that some of the other Honeycomb libraries are capable of and it is really helpful!

### Discussion
Moved all sampling behaviour to `finishSpan()`, following the pattern demonstrated in the beelines-python library ([see here]). This will allow users to define custom sampler hooks that meet a standard interface. This change required refactor of existing deterministic sampler (and its tests) to fit with the new interface requirements.

### What's missing?
1. I haven't added any documentation. Usage docs don't appear to live in this repo.
2. There are no tests for what happens if you do not provide a samplerHook or a sampleRate. See roughly line 224 in `lib/api/libhoney.js` for more detail. I wasn't sure the best way to verify that `send()` was called rather than `sendPresampled()`.

[see here]: https://github.com/honeycombio/beeline-python/blob/master/beeline/__init__.py#L179

### DIFF
--- a/lib/api/libhoney.js
+++ b/lib/api/libhoney.js
@@ -222,12 +222,13 @@ module.exports = class LibhoneyEventAPI {
       });
 
       let presampled = false;
+      const eventDataForLogs = JSON.stringify(ev.data);
       if (this.samplerHook) {
-        debug(`executing sampler hook on event ev = ${JSON.stringify(ev.data)}`);
+        debug(`executing sampler hook on event ev = ${eventDataForLogs}`);
         const samplerHookResponse = this.samplerHook(ev.data);
         const keepEvent = samplerHookResponse.shouldSample;
         if (!keepEvent) {
-          debug("skipping event due to sampler hook sample ev = " + JSON.stringify(ev.data));
+          debug(`skipping event due to sampler hook sample ev = ${eventDataForLogs}`);
           return;
         }
 
@@ -239,12 +240,12 @@ module.exports = class LibhoneyEventAPI {
       }
 
       if (this.presendHook) {
-        debug(`executing presend hook on event ev = ${JSON.stringify(ev.data)}`);
+        debug(`executing presend hook on event ev = ${eventDataForLogs}`);
         this.presendHook(ev.data);
       }
 
       if (presampled) {
-        debug("enqueuing presampled event ev = " + JSON.stringify(ev.data));
+        debug(`enqueuing presampled event ev = ${eventDataForLogs}`);
         ev.sendPresampled();
       }
     });
--- a/lib/api/libhoney.js
+++ b/lib/api/libhoney.js
@@ -232,8 +232,7 @@ module.exports = class LibhoneyEventAPI {
           return;
         }
 
-        const sampleRate = samplerHookResponse.sampleRate;
-        if (typeof sampleRate === "number") {
+        if (typeof samplerHookResponse.sampleRate === "number") {
           ev.sampleRate = samplerHookResponse.sampleRate;
         }
         presampled = true;
--- a/lib/api/libhoney.js
+++ b/lib/api/libhoney.js
@@ -237,8 +237,7 @@ module.exports = class LibhoneyEventAPI {
       }
 
       if (this.presendHook) {
-        debug(`executing presend hook on event ev = ${eventDataForLogs}`);
-        this.presendHook(ev.data);
+        this.presendHook(ev);
       }
 
       debug(`enqueuing presampled event ev = ${eventDataForLogs}`);
--- a/lib/api/libhoney.js
+++ b/lib/api/libhoney.js
@@ -216,6 +216,7 @@ module.exports = class LibhoneyEventAPI {
 
       let presampled = false;
       if (this.samplerHook) {
+        debug(`executing sampler hook on event ev = ${JSON.stringify(ev.data)}`);
         const samplerHookResponse = this.samplerHook(ev.data);
         const keepEvent = samplerHookResponse.shouldSample;
         if (!keepEvent) {
--- a/lib/api/libhoney.js
+++ b/lib/api/libhoney.js
@@ -231,9 +231,9 @@ module.exports = class LibhoneyEventAPI {
           return;
         }
 
-        const newSampleRate = samplerHookResponse.newSampleRate;
-        if (newSampleRate !== undefined && typeof newSampleRate === "number") {
-          ev.sampleRate = samplerHookResponse.newSampleRate;
+        const sampleRate = samplerHookResponse.sampleRate;
+        if (typeof sampleRate === "number") {
+          ev.sampleRate = samplerHookResponse.sampleRate;
         }
         presampled = true;
       }
--- a/lib/api/libhoney.js
+++ b/lib/api/libhoney.js
@@ -30,15 +30,15 @@ const getFilteredOptions = opts => {
 
 module.exports = class LibhoneyEventAPI {
   constructor(opts) {
-    if (typeof opts["samplerHook"] === "function") {
-      this.samplerHook = opts["samplerHook"];
+    if (typeof opts.samplerHook === "function") {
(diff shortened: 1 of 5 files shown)


lib/api/libhoney.js
```diff
@@ -205,10 +201,30 @@ module.exports = class LibhoneyEventAPI {
...
+        const newSampleRate = samplerHookResponse.newSampleRate;
+        if (newSampleRate !== undefined && typeof newSampleRate === "number") {
+          ev.sampleRate = samplerHookResponse.newSampleRate;
+        }
+        presampled = true;
+      }
+
+      if (presampled) {
+        debug("enqueuing presampled event ev = " + JSON.stringify(ev.data));
+        ev.sendPresampled();
```
toshok: always use `sendPresampled` here.  we configure the libhoney instance we're using to not do any further sampling, so `ev.send()` in the other branch isn't necessary.

lib/api/libhoney.js
```diff
@@ -70,11 +71,6 @@ module.exports = class LibhoneyEventAPI {
   startTrace(metadataContext, traceId, parentSpanId) {
     const id = traceId || uuidv4();
 
-    if (this.ds && !this.ds.sample(id)) {
```
toshok: this makes me quite sad.  a bunch of wasted work/cpu cycles when we know we're not going to send any events in the trace at all.

lib/api/libhoney.js
```diff
@@ -58,7 +59,7 @@ module.exports = class LibhoneyEventAPI {
           dataset: process.env["HONEYCOMB_DATASET"] || defaultName,
           userAgentAddition: `honeycomb-beeline/${pkg.version}`,
         },
-        optsWithoutSampleRate
+        opts
```
toshok: we shouldn't be passing the sample rate to libhoney.  any sampling we're doing is done here already.  there's no reason to include libhoney in it.

lib/api/libhoney.js
```diff
@@ -20,19 +20,20 @@ const incr = (payload, key, val = 1) => {
...
   constructor(opts) {
-    let sampleRate;
-    let optsWithoutSampleRate = {};
-    Object.keys(opts).forEach(k => {
-      if (k === "sampleRate") {
-        sampleRate = opts[k];
-      } else {
-        optsWithoutSampleRate[k] = opts[k];
+    if ("samplerHook" in opts) {
+      this.samplerHook = opts["samplerHook"];
```
toshok:
```suggestion
      this.samplerHook = opts.samplerHook;
```

lib/api/libhoney.js
```diff
@@ -20,19 +20,20 @@ const incr = (payload, key, val = 1) => {
...
+      debug("using custom samplerHook");
+    }
+
+    if (!("samplerHook" in opts)) {
+      const sampleRate = opts["sampleRate"];
```
toshok:
```suggestion
      const sampleRate = opts.sampleRate;
```

lib/api/libhoney.js
```diff
@@ -205,10 +201,30 @@ module.exports = class LibhoneyEventAPI {
...
+      let presampled = false;
+      if (this.samplerHook) {
+        const samplerHookResponse = this.samplerHook(ev.data);
+        const keepEvent = samplerHookResponse.shouldSample;
+        if (!keepEvent) {
+          debug("skipping event due to sampler hook sample ev = " + JSON.stringify(ev.data));
+          return;
+        }
+
+        const newSampleRate = samplerHookResponse.newSampleRate;
```
toshok: s/`newSampleRate`/`sampleRate`/ everywhere, imo, but especially in the sampler hook response

lib/deterministic_sampler.test.js
```diff
@@ -43,11 +64,14 @@ test("it works (stastically)", () => {
   const testSampleRates = [1, 2, 10];
 
   for (let sampleRate of testSampleRates) {
-    let ds = new DeterministicSampler(sampleRate);
     let nSampled = 0;
 
     for (let i = 0; i < nRequestIDs; i++) {
-      if (ds.sample(randomRequestID())) {
+      const testSampler = deterministicSampler(sampleRate);
```
toshok: not a huge deal, but move outside this for loop.

lib/api/libhoney.js
```diff
@@ -205,10 +201,30 @@ module.exports = class LibhoneyEventAPI {
...
+        if (newSampleRate !== undefined && typeof newSampleRate === "number") {
```
toshok: can remove the `undefined` check here
```suggestion
        if (typeof newSampleRate === "number") {
```

lib/api/libhoney.js
```diff
@@ -205,10 +201,30 @@ module.exports = class LibhoneyEventAPI {
...
+          ev.sampleRate = samplerHookResponse.newSampleRate;
+        }
+        presampled = true;
+      }
+
+      if (presampled) {
+        debug("enqueuing presampled event ev = " + JSON.stringify(ev.data));
```
toshok: I don't think we need to stringify the span data twice per `finishSpan`.  Either remove the stringify here or from the top of this function.

we're also using new enough nodes that we should use interpolated strings here:
```suggestion
        debug(`enqueuing event ${JSON.stringify(ev)}`);
```
(^ note `ev` instead of `ev.data` so that it also logs the sample rate.)

lib/api/libhoney.js
allyjweir: Hopefully addressed with commit 7df9351d2db6a1f88016e5ea08dbdcdc2d54697e

allyjweir:
Thanks for the thorough review @toshok. I'm on GMT (just wrapping my day up, 6:30pm here) so I will address these things tomorrow morning and hopefully will be ready for you to take a look again.

As you can see, I've pushed up the addition of a `presendHook` implementation as well. This was something I needed to add and thought it aligned closely with the existing PR and easily reviewed together.

lib/api/libhoney.js
allyjweir: We could add this back in. If we put the fallback `deterministicSampler` under a different key (i.e. if it didn't get assigned to `this.samplerHook` then we could check for existance and call it here.)

It might be a bit confusing that we have radically different approaches to sampling depending on if you provide your own custom hook or not.

lib/api/libhoney.js
allyjweir: Cool cool. I usually work in Babel so my 'definition' of node is fairly confused.

allyjweir: What I've done for now is extract the `stringify` action itself to a variable so at least we aren't re-stringifying every time.

I copied the logging strategy from the `beeline-python` library. Had found having this much debug very helpful when getting up and running. Given these logs will only be called if the debug mode is enabled, could we possibly keep them in?

lib/api/libhoney.js
allyjweir: Just to confirm, if no samplerHook or sampleRate provided, we should send every event?

toshok: yep, that's expected behavior

allyjweir:
@toshok What further work needs done to get this across the line and merged?

lib/api/libhoney.js
allyjweir: Cool cool, I've added a test to the test suite that checks this.

lib/api/libhoney.js
toshok: nah, it's only going to be useful if other beelines pick it up as well.  we also need to worry about communicating the sampling decision through propagated headers even if we aren't sampling.

toshok:
nothing! :) 🎉

nathanleclaire:
Thank you @allyjweir !

PR #129, opened 2019-03-12
Parameterise startSpan to accept custom span and parent ids

allyjweir:
While developing an Apollo Server extension to trace the resolving of fields in a GraphQL query, it became necessary to be able to modify a new span’s parent due to the asynchronous nature of these resolvers.

### DIFF
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -60,8 +60,8 @@ module.exports = {
     }
   },
 
-  startSpan(metadataContext) {
-    return apiImpl.startSpan(metadataContext);
+  startSpan(metadataContext, spanId = undefined, parentId = undefined) {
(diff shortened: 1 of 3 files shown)


toshok:
hm..  I don't suppose there's any chance of seeing a motivating example for this?  I'm on the fence about exposing the ability to specify parent ids like this in the public api.

allyjweir:
I'm also on the fence about it!

Here is a link to the code motivating this change. Specifically see line 122: https://gist.github.com/allyjweir/a32288be348bc99df009d478201344b7

allyjweir:
Backing away from this one for now. The GraphQL extension I needed this for was good, too good. Overloaded our Honeycomb instance with events and we disabled it. For now, pausing work and reconsidering a different approach so this code is no longer required.

PR #130, opened 2019-03-14
Wrap Bluebird _then method, all promise fns go through here

tgriesser:
When using Bluebird, all promise helpers / methods (`.map`, `.race`, `.tap`, `.reflect`, etc.) are passed through an internal [`_then` method](https://github.com/petkaantonov/bluebird/blob/49da1ac256c7ee0fb1e07679791399f24648b933/src/promise.js#L220-L277). This PR wraps that `_then` rather than the top level `then` to ensure the trace fns are bound when called from these helpers.


MikeGoldsmith:

PR #134, opened 2019-03-31
fix update function calls to use fourArgs

thegrumpysnail:
`replaceOne`, `update`, `updateOne` and `updateMany` take four arguments, not three - refer to https://mongodb.github.io/node-mongodb-native/3.2/api/Collection.html#update for example.

As is, using three arguments + promise method (no callback) results in ...

```
TypeError: callback is not a function
.../node_modules/honeycomb-beeline/lib/instrumentation/mongodb.js:59:24
```


toshok:
👍

PR #137, opened 2019-04-09
[mongodb] handle null options

### LINKED ISSUE #136: Mongoose passing `null` options crashes instrumentation in prefixKeys
In what may have been a recent update, Mongoose is passing `null` into the MongoDB API for the options to `insertOne` and others, which is tripping up the Beeline instrumentation.

You can see the troublesome Mongoose code here: https://github.com/Automattic/mongoose/blob/5.4.22/lib/model.js#L242

This `null` then makes it way into `prefixKeys` and crashes here: https://github.com/honeycombio/beeline-nodejs/blob/master/lib/instrumentation/mongodb.js#L8

I have fixed this locally by changing `Object.keys(map)` to `Object.keys(map || {})` but there may be a more appropriate fix.

toshok:
Fixes #136 

Change `prefixKeys` from defaulting the `map` parameter to `{}` (which only works when the argument is `undefined`) and instead guard around the `addContext("db.options", options)` call.

This means that `additionalContext` callbacks _must_ return an object if present (which they all do).

### DIFF
--- a/lib/instrumentation/mongodb.js
+++ b/lib/instrumentation/mongodb.js
@@ -5,9 +5,11 @@ const shimmer = require("shimmer"),
 
 function prefixKeys(prefix, map) {
   let prefixed = {};
-  Object.keys(map).forEach(k => {
-    prefixed[`${prefix}.${k}`] = map[k];
-  });
+  if (map) {
+    Object.keys(map).forEach(k => {
+      prefixed[`${prefix}.${k}`] = map[k];
+    });
+  }
   return prefixed;
 }
 
--- a/lib/instrumentation/mongodb.js
+++ b/lib/instrumentation/mongodb.js
@@ -3,7 +3,7 @@ const shimmer = require("shimmer"),
   api = require("../api"),
   schema = require("../schema");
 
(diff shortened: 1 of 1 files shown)


lib/instrumentation/mongodb.js
```diff
@@ -48,7 +48,9 @@ function instrumentMongodb(mongodb, opts = {}) {
             [schema.TRACE_SPAN_NAME]: eventName,
           },
           span => {
-            api.addContext(prefixKeys("db.options", options));
+            if (options) {
```
toshok: this and line 6 above are the only code changes.  the rest are thanks to `prettier`.

SeanHood:
Hi, wondering if any more progress has been made on this issue? We have had to disable mongodb instrumentation for now in our app. Loving Honeycomb for everything else though 👌

toshok:
> makes sense to me.
> 
> Did you consider adding the guard in `prefixKeys`? wrap it in an `if(map)` then checking the output before adding calling `addContext`? It seems like it might be a safer option for callers of `prefixKeys`

done

PR #140, opened 2019-05-09
switch from travis-ci to circleci

toshok:
pretty much identical to the libhoney-js conversion, just with different node versions since we depend on `async_hooks` (and therefore need a newer node version.)


README.md
```diff
@@ -1,15 +1,15 @@
 # Honeycomb Beeline for NodeJS
 
-[![Build Status](https://travis-ci.org/honeycombio/beeline-nodejs.svg?branch=master)](https://travis-ci.org/honeycombio/beeline-nodejs)
+[![CircleCI](https://circleci.com/gh/honeycombio/beeline-nodejs.svg?style=svg)](https://circleci.com/gh/honeycombio/beeline-nodejs)
```
maplebed: style=shield?

toshok: oh yeah, I was wondering how to specify that.  will fix.

PR #146, opened 2019-06-21
dataset propagation, fix parent span ids on async spans

toshok:
The go beeline will put a dataset k/v in the marshaled trace context.  If it's there, use it.  This will result in distributed traces being sent entirely to the same dataset (chosen by the root service.)

The async span problem is a little difficult to explain: The node beeline creates async spans by essentially forking/execing the trace context, putting the new span as the root span of the new trace context.  everything else is copied over, so the trace id remains the same, etc.  The parent id is also grabbed so the root async span will be shown as a child of the current span at the time the async span was created.

The problem comes in when that root async span finishes, and another span starts.  The stack is now empty, so there's no parentId to use.  The fix is to also store the parent id on the new trace context, and use it from there if the stack is empty.

Lastly, I noticed that in the express instrumentation when unmarshaling trace context, we were using the customContext api to add the context from the calling service.  This results in `app.` being prepended, so we were seeing fields like `app.app.user_id`, `app.app.whatever`.

### DIFF
--- a/lib/instrumentation/express.js
+++ b/lib/instrumentation/express.js
@@ -191,13 +191,12 @@ const getMagicMiddleware = ({ userContext, traceIdSource, packageVersion }) => (
       "request.xhr": req.xhr,
     },
     traceContext.traceId,
-    traceContext.parentSpanId
+    traceContext.parentSpanId,
+    traceContext.dataset
   );
 
   if (traceContext.customContext) {
-    Object.keys(traceContext.customContext).forEach(k =>
-      api.customContext.add(k, traceContext.customContext[k])
-    );
+    api.addContext(traceContext.customContext);
   }
 
   if (!trace) {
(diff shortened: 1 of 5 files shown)


lib/instrumentation/express.js
```diff
@@ -191,13 +191,12 @@ const getMagicMiddleware = ({ userContext, traceIdSource, packageVersion }) => (
...
-    traceContext.parentSpanId
+    traceContext.parentSpanId,
+    traceContext.dataset
   );
 
   if (traceContext.customContext) {
-    Object.keys(traceContext.customContext).forEach(k =>
-      api.customContext.add(k, traceContext.customContext[k])
-    );
+    api.addContext(traceContext.customContext);
```
toshok: this is the fix for `app.app.*` from the summary

PR #153, opened 2019-07-19
Add support for mysql2#createPool

ashishbista:
Before this, you wouldn’t get mysql queries in the traces if you were using `mysql2#createPool` instead of `mysql2#createConnection`. This commit adds support for `mysql2#createPool`.

More on mysql2#createPool: https://github.com/sidorares/node-mysql2#using-connection-pools

The following sample app can be used to test it:

```
# Schema
CREATE TABLE `posts` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `title` varchar(255) DEFAULT NULL,
  `body` text,
  `created_at` datetime NOT NULL,
  `updated_at` datetime NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=28 DEFAULT CHARSET=utf8;
```

```
const beeline =  require("honeycomb-beeline")({
 writeKey: “xxxxxxxx”,
 dataset: “xxxxxx”,
 serviceName: "my-awesome-service"
});

const express = require('express')
const app = express()
const port = 3000

const mysql = require('mysql2');

const connection = mysql.createConnection({
  host: 'localhost',
  user: 'root',
  password: ‘xxxxx’,
  database: 'awesome_development'
});

const pool = mysql.createPool({
  host: '127.0.0.1',
  user: 'root',
  password: ‘xxxxx’,
  database: 'awesome_development'
});

app.get('/', (req, res) => {
  pool.query(
    'SELECT * FROM `posts` WHERE `title` = "pool.query"',
    function(err, results, fields) {
    }
  );

  pool.execute(
    'SELECT * FROM `posts` WHERE `title` = "pool.execute"',
    function(err, results, fields) {
    }
  );

  connection.query(
    'SELECT * FROM `posts` WHERE `title` = "connection.query"',
    function(err, results, fields) {
    }
  );

  connection.execute(
    'SELECT * FROM `posts` WHERE `title` = "connection.execute"',
    function(err, results, fields) {
    }
  );

  pool.getConnection(function(err, poolConnection1) {
    poolConnection1.query(
      'SELECT * FROM `posts` WHERE `title` = "poolConnection1.query"',
      function(err, results, fields) {
      }
    );
  });

  pool.getConnection(function(err, poolConnection1) {
    poolConnection1.execute(
      'SELECT * FROM `posts` WHERE `title` = "poolConnection2.execute"',
      function(err, results, fields) {
      }
    );
  });

  pool.getConnection(function(err, poolConnection2) {
    poolConnection2.query(
      'SELECT * FROM `posts` WHERE `title` = "poolConnection2.query"',
      function(err, results, fields) {
      }
    );
  });
  pool.getConnection(function(err, poolConnection2) {
    poolConnection2.execute(
      'SELECT * FROM `posts` WHERE `title` = "poolConnection2.execute"',
      function(err, results, fields) {
      }
    );
  });

  res.send({});
})

```

### DIFF
--- a/lib/instrumentation/mysql2.js
+++ b/lib/instrumentation/mysql2.js
@@ -3,49 +3,42 @@ const shimmer = require("shimmer"),
   api = require("../api"),
   schema = require("../schema");
 
-let instrumentConnection = function(conn, packageVersion) {
-  shimmer.massWrap(conn, ['query', 'execute'], function(original) {
+let instrumentMysql2 = function(mysql2, opts = {}) {
+  shimmer.massWrap(mysql2.Connection.prototype, ["execute", "query"], function(original) {
     return function(...args) {
-      if (args.length < 1) {
+      if (args.length < 1 || !api.traceActive()) {
         return original.apply(this, args);
       }
-      if (!api.traceActive()) {
-        return original.apply(this, args);
+
+      query = args[0].sql;
+      if (typeof query === "undefined") {
+        query = args[0];
       }
-      query = args[0].sql
-      if (typeof query === 'undefined') {
-        query = args[0]
-      }
-      api.startAsyncSpan(
+      return api.startAsyncSpan(
         {
           [schema.EVENT_TYPE]: "mysql2",
-          [schema.PACKAGE_VERSION]: packageVersion,
+          [schema.PACKAGE_VERSION]: opts.packageVersion,
           [schema.TRACE_SPAN_NAME]: "query",
           "db.query": query,
         },
         span => {
-          let cb = args[args.length - 1];
+          let callback = args[args.length - 1];
           // XXX(toshok) this bindFunction shouldn't be necessary, but we aren't
           // finishing the event properly without it.
-          let wrapped_cb = api.bindFunctionToTrace(function(...cb_args) {
+          if (callback && typeof callback === "function") {
+            let wrapped_callback = api.bindFunctionToTrace(function(...callback_args) {
+              api.finishSpan(span, "query");
+              return callback(...callback_args);
+            });
+            return original.apply(this, args.slice(0, -1).concat(wrapped_callback));
+          }
+          try {
+            return original.apply(this, args);
+          } finally {
             api.finishSpan(span, "query");
-            return cb(...cb_args);
-          });
-
-          return original.apply(this, args.slice(0, -1).concat(wrapped_cb));
+          }
         }
       );
-    };
-  });
-
-  return conn;
-};
-
-let instrumentMysql2 = function(mysql2, opts = {}) {
-  shimmer.wrap(mysql2, "createConnection", function(original) {
-    return function(...args) {
-      let conn = original.apply(this, args);
-      return instrumentConnection(conn, opts.packageVersion);
     };
   });
   return mysql2;


toshok:
I worry that the pool is creating some number of connections internally at initialization time that are benefiting from the shim propagating context, and that the context won't make it to additional connections created.  Is that a valid concern, do you think?

any chance you could make sure the following pool+connection usage works in your sample app?

```
const pool = mysql.createPool({ ... });

pool.getConnection(function(err, conn) {
  conn.query(/* the query you have above */);

  pool.getConnection(function(err, conn2) {
    conn.execute(/* the execute you have above */);
    // don't release conn2
  });

  // don't release conn
});
```

ashishbista:
@toshok that's totally a valid concern.

My previous commit didn't work with the mix of pool and connection. I had to update the commit to make it work with pool+connection as suggested by you.

ashishbista:
Speaking of `query` call, I found three classes calling the`query` function.

- `mysql2.Connection.query`
- `mysql2.PoolConnection.query`
- `mysql2.Pool.query`

[mysql2.Connection#query](https://github.com/sidorares/node-mysql2/blob/master/lib/connection.js#L483) is the ultimate gateway function that queries MySQL.

`mysql.PoolConnection` class doesn't have `query` function rather it inherits the same function from its parent `mysql2.Connection`, look at https://github.com/sidorares/node-mysql2/blob/master/lib/pool_connection.js#L5.

`mysql2.Pool` delegates query call to `mysql2.PoolConnection`, look at https://github.com/sidorares/node-mysql2/blob/master/lib/pool.js#L145

So, I think monkey patching `mysql2.Connection.prototype#query` should cover all query calls.

Same for the `execute` function.

cc @toshok

toshok:
> So, I think monkey patching mysql2.Connection.prototype#query should cover all query calls.
>
> Same for the execute function.

awesome

lib/instrumentation/mysql2.js
```diff
@@ -3,49 +3,39 @@ const shimmer = require("shimmer"),
...
-            return cb(...cb_args);
-          });
-
-          return original.apply(this, args.slice(0, -1).concat(wrapped_cb));
+          if (callback && typeof callback === 'function') {
+            let wrapped_callback = api.bindFunctionToTrace(function(...callback_args) {
+              api.finishSpan(span, 'query');
+              return callback(...callback_args);
+            });
+            return original.apply(this, args.slice(0, -1).concat(wrapped_callback));
```
toshok: hm, without commit history I can't tell if this was something that snuck in only recently or if it's always been there and I missed it, but this will only call the original function if there's a callback present.  we should always call through to the original, but the argument list should be modified only if we wrapped the callback.

otherwise lgtm.

ashishbista: @toshok Sorry for not running the linter. I missed it because I use a git gui tool.

I added `return original.apply(this, args);` after this line. Should it fix that?

lib/instrumentation/mysql2.js
```diff
@@ -3,49 +3,40 @@ const shimmer = require("shimmer"),
...
-          return original.apply(this, args.slice(0, -1).concat(wrapped_cb));
+          if (callback && typeof callback === "function") {
+            let wrapped_callback = api.bindFunctionToTrace(function(...callback_args) {
+              api.finishSpan(span, "query");
+              return callback(...callback_args);
+            });
+            return original.apply(this, args.slice(0, -1).concat(wrapped_callback));
+          }
+          api.finishSpan(span, "query");
+          return original.apply(this, args);
```
toshok: this won't include the runtime of the function itself in the span.   Suggest:

```
try {
    return original.apply(this, args);
} finally {
    api.finishSpan(span, "query");
}
```

lib/instrumentation/mysql2.js
```diff
@@ -3,49 +3,40 @@ const shimmer = require("shimmer"),
...
-
-  return conn;
-};
-
-let instrumentMysql2 = function(mysql2, opts = {}) {
-  shimmer.wrap(mysql2, "createConnection", function(original) {
-    return function(...args) {
-      let conn = original.apply(this, args);
-      return instrumentConnection(conn, opts.packageVersion);
+      return original.apply(this, args);
```
toshok: this one isn't necessary (and will break things, since it'll run `original` twice.)

ashishbista: I was thinking that as well. But, my sample app crashes without it:

```
WHERE `title` = \"pool.query\"","trace.trace_id":"69c9489c-58e1-4c38-bf6e-b7276fe7002f","trace.span_id":"ff511718-85a8-48a2-9332-8ce8ddefbc2a","duration_ms":0.22707,"response.status_code":"304","meta.instrumentations":["express","http","mysql2"],"meta.instrumentation_count":3,"meta.beeline_version":"1.5.1","meta.node_version":"v10.16.0"} +0ms
/Users/bistaa/Work/services/my-awesome-service/node_modules/mysql2/lib/pool.js:145
      conn.query(cmdQuery).once('end', () => {
                          ^

TypeError: Cannot read property 'once' of undefined
    at getConnection (/Users/bistaa/Work/services/my-awesome-service/node_modules/mysql2/lib/pool.js:145:27)
    at connection.connect.err (/Users/bistaa/Work/services/my-awesome-service/node_modules/mysql2/lib/pool.js:63:16)
    at PoolConnection.<anonymous> (/Users/bistaa/Work/services/my-awesome-service/node_modules/mysql2/lib/connection.js:759:13)
```

The sample app works perfectly with this line.

lib/instrumentation/mysql2.js
ashishbista: Added this

PR #154, opened 2019-07-19
initial fastify instrumentation

toshok:
Introduce `fastify` auto-instrumentation.  It registers two hooks before anything in user-code is run, one `onRequest` to start the trace, and one `onResponse` to send it.  Given just how pervasively async the api is and how things are registered, it's not as easy to ensure that context gets propagated correctly through hooks, so instead we store off the trace context in a Map and fetch it for every hook.  This means that there may be blips where context is lost, but each hook/handler should begin anew with the context.

Things this PR probably needs to include:

- [x] hooks
- [x] route handlers (`.get`, `.put`, etc)
- [ ] middleware - skipping this for now.  if we have requests, can add it via another PR.

### DIFF
--- a/lib/instrumentation/fastify.js
+++ b/lib/instrumentation/fastify.js
@@ -1,0 +1,36 @@
+/* eslint-env node */
+// const onHeaders = require("on-headers"),
+// shimmer = require("shimmer"),
+// flatten = require("array-flatten"),
+// tracker = require("../async_tracker"),
+// schema = require("../schema"),
+// api = require("../api"),
+// util = require("../util"),
+// path = require("path"),
+// pkg = require(path.join(__dirname, "..", "..", "package.json")),
+// traceUtil = require("trace-util"),
+// debug = require("debug")(`${pkg.name}:fastify`);
+
+let instrumentFastify = function(fastify, _opts = {}) {
+  // let userContext, traceIdSource;
+  // if (opts.userContext) {
+  //   if (Array.isArray(opts.userContext) || typeof opts.userContext === "function") {
+  //     userContext = opts.userContext;
+  //   } else {
+  //     debug(
+  //       "userContext option must either be an array of field names or a function returning an object"
+  //     );
+  //   }
+  // }
+  // if (opts.traceIdSource) {
+  //   if (typeof opts.traceIdSource === "string" || typeof opts.traceIdSource === "function") {
+  //     traceIdSource = opts.traceIdSource;
+  //   } else {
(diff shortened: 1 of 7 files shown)


lib/instrumentation/fastify.js
```diff
@@ -0,0 +1,228 @@
...
+
+      finishersByRequest.set(request, boundFinisher);
+      trackedByRequest.set(request, tracker.getTracked());
+
+      next();
+    });
+
+    app.addHook("onResponse", (request, reply, next) => {
+      // calculate total time for the request and send the root span
+      debug("in onResponse");
```
dsoo: Do we want to leave the various debug spam in?

lib/instrumentation/fastify.js
```diff
@@ -0,0 +1,228 @@
...
+      const finisher = finishersByRequest.get(request);
+      if (finisher) {
+        debug("finishing");
+        finisher(reply);
+      }
+
+      next();
+    });
+
+    // now that we've added our special hooks, instrument the addHook method such that
```
dsoo: Incomplete comment?

lib/instrumentation/fastify.js
```diff
@@ -0,0 +1,228 @@
...
+          let request = args[0];
+
+          let tracked = trackedByRequest.get(request);
+          tracker.setTracked(tracked);
+
+          let span = api.startSpan({
+            name: `${hookName} hook`,
+            [schema.EVENT_TYPE]: "fastify",
+            [schema.PACKAGE_VERSION]: opts.packageVersion,
+            // XXX toshok more here
```
dsoo: ???

lib/instrumentation/fastify.js
```diff
@@ -0,0 +1,228 @@
...
+
+        let wrappedHandler = function(request, reply) {
+          let tracked = trackedByRequest.get(request);
+          tracker.setTracked(tracked);
+
+          let span = api.startSpan({
+            name: `handler: ${method} ${handlerUrl}`,
+            [schema.EVENT_TYPE]: "fastify",
+            [schema.PACKAGE_VERSION]: opts.packageVersion,
+            // XXX toshok more here
```
dsoo: ???

lib/instrumentation/fastify.js
toshok: ah, definitely not. 👍

wolfeidau:
Is there any examples of how to use this?

I can't find it in the documentation.

Issue #37, opened 2018-03-03
pg instrumentation


toshok:
not everyone uses mysql :)

vreynolds:
pg instrumentation has been added, if something is missing please open a new issue

PR #155, opened 2019-07-24
Add instrumentation for `pg` package

allyjweir:
Addresses issue #37.

Our team uses the [pg package](https://www.npmjs.com/package/pg) to manage our Postgres interactions and were looking to extend our tracing to include those queries to the database. Didn't see an easy way to introduce a custom instrumentation from our own codebase so looking to upstream and share.

A lot of this is based off the pre-existing mysql2 (thanks to @toshok and @ashishbista for their work on that one!) and also a little inspiration from [opentracing-auto's pg instrumentation](https://github.com/RisingStack/opentracing-auto/blob/master/src/instrumentation/pg.js).

Main difference to the mysql2 instrumentation is that this enhances the span with error/result information, similarly to [how beeline-python does it.](https://github.com/honeycombio/beeline-python/blob/master/beeline/middleware/django/__init__.py#L17)

[Here is a minimum project to test the instrumentation out with](https://github.com/allyjweir/pg-trace-example). Check the README.md for setup details.

### DIFF
--- a/lib/instrumentation/pg.js
+++ b/lib/instrumentation/pg.js
@@ -3,6 +3,23 @@ const shimmer = require("shimmer"),
   api = require("../api"),
   schema = require("../schema");
 
+const getQueryString = function(args) {
+  if (typeof args[0] === "string") return args[0];
+  if (typeof args[0] === "object") return args[0].text;
+  return undefined;
+};
+
+const getQueryArgs = function(args) {
+  if (typeof args[0] === "object") return args[0].values;
+  if (Array.isArray(args[1])) return args[1];
+  return undefined;
+};
+
+const getQueryName = function(args) {
+  if (typeof args[0] === "object") return args[0].name;
+  return undefined;
+};
+
 let instrumentPg = function(pg, opts = {}) {
   shimmer.wrap(pg.Client.prototype, "query", function(query) {
     return function(...args) {
@@ -13,17 +30,14 @@ let instrumentPg = function(pg, opts = {}) {
         return query.apply(this, args);
       }
 
-      let queryString = args[0].text;
-      if (typeof queryString === "undefined") {
-        queryString = args[0];
-      }
       api.startAsyncSpan(
         {
           [schema.EVENT_TYPE]: "pg",
           [schema.PACKAGE_VERSION]: opts.packageVersion,
           [schema.TRACE_SPAN_NAME]: "query",
-          "db.query": queryString,
-          "db.query_args": args[1],
+          "db.query": getQueryString(args),
+          "db.query_args": getQueryArgs(args),
+          "db.query_name": getQueryName(args),
         },
         span => {
           let cb = args[args.length - 1];
--- a/lib/instrumentation/pg.js
+++ b/lib/instrumentation/pg.js
@@ -1,0 +1,62 @@
+/* eslint-env node */
+const shimmer = require("shimmer"),
+  api = require("../api"),
+  schema = require("../schema");
+
+let instrumentPg = function(pg, opts = {}) {
+  shimmer.wrap(pg.Client.prototype, "query", function(query) {
+    return function(...args) {
+      if (args.length < 1) {
+        return query.apply(this, args);
+      }
+      if (!api.traceActive()) {
+        return query.apply(this, args);
+      }
+
+      let queryString = args[0].text;
+      if (typeof queryString === "undefined") {
+        queryString = args[0];
+      }
+      api.startAsyncSpan(
+        {
+          [schema.EVENT_TYPE]: "pg",
+          [schema.PACKAGE_VERSION]: opts.packageVersion,
+          [schema.TRACE_SPAN_NAME]: "query",
+          "db.query": queryString,
+          "db.query_args": args[1],
+        },
+        span => {
+          let cb = args[args.length - 1];
+          // XXX(toshok) this bindFunction shouldn't be necessary, but we aren't
+          // finishing the event properly without it.
+          let wrapped_cb = api.bindFunctionToTrace(function(...cb_args) {
+            let error = cb_args[0];
+            let result = cb_args[1];
+
+            if (error !== null && error instanceof Error) {
+              api.addContext({
+                "db.error": error.message,
+                "db.error_stack": error.stack,
+                "db.error_hint": error.hint,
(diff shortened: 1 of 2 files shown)


lib/instrumentation/pg.js
```diff
@@ -0,0 +1,62 @@
...
+        {
+          [schema.EVENT_TYPE]: "pg",
+          [schema.PACKAGE_VERSION]: opts.packageVersion,
+          [schema.TRACE_SPAN_NAME]: "query",
+          "db.query": queryString,
+          "db.query_args": args[1],
+        },
+        span => {
+          let cb = args[args.length - 1];
+          // XXX(toshok) this bindFunction shouldn't be necessary, but we aren't
```
toshok: Is this true for `pg` as well?  if so we should come up with a better comment :)

lib/instrumentation/pg.js
```diff
@@ -0,0 +1,62 @@
...
+      if (typeof queryString === "undefined") {
+        queryString = args[0];
+      }
+      api.startAsyncSpan(
+        {
+          [schema.EVENT_TYPE]: "pg",
+          [schema.PACKAGE_VERSION]: opts.packageVersion,
+          [schema.TRACE_SPAN_NAME]: "query",
+          "db.query": queryString,
+          "db.query_args": args[1],
```
toshok: maybe this is a failing in the mysql instrumentation too, but looking at the pg docs, if `args[0]` is an object, the query args will be `args[0].values`, and `args[1]` will be the callback, right?

(sidenote: again maybe mysql2 instrumentation is lacking, but there's also some other great context to be had in the object-as-first-arg in pg.  `query.name`, `query.rowMode`.)

lib/instrumentation/pg.js
```diff
@@ -0,0 +1,62 @@
...
+              api.addContext({
+                "db.rows_affected": result.rowCount,
+              });
+            }
+
+            api.finishSpan(span, "query");
+            return cb(...cb_args);
+          });
+
+          return query.apply(this, args.slice(0, -1).concat(wrapped_cb));
```
toshok: this is a more complicated API, since you can omit the callback and `pg` will return a promise :/

I ran into a similar API for `fastify` instrumentation: see https://github.com/honeycombio/beeline-nodejs/pull/154/files#diff-e5ed2aa7f0ba33ce9e2ecbe84645fa65R138-R166

steps there are:

1. check if last arg is a callback.  if it is, wrap it and modify the args array.
2. call the original method
3. if the return value is not a promise, we're done - return it.
4. otherwise, return a new promise that finishes the span in the `.then` method and chains resolves/rejects.

Hm, I suppose that recipe allows for a function that both accepts a callback _and_ returns a promise, both of which finish the span..  open to a more restrictive recipe that disallows that (probably calling/returning in 1 above ?)

lib/instrumentation/pg.js
allyjweir: This was a lift and shift from the mysql2 instrumentation. Will re-examine if this is necessary.

lib/instrumentation/pg.js
allyjweir: Yep, I had only tested for the 'parameterised query' model arguments pattern described [here](https://node-postgres.com/features/queries).

What do you think would be the best way to handle the fact `.query()` can be called 3 ways?

For reference:
1. `.query(queryString, callback)`
2. `.query(queryString, values, callback)`
3. `.query(queryObject, callback)`

Perhaps look at how the library itself distinguishes between them itself and then use that to gather as much context to add to the span as possible?

allyjweir: Of interest: https://github.com/brianc/node-postgres/blob/master/lib/utils.js#L141

lib/instrumentation/pg.js
allyjweir: Interestingly, the [example service I used to test this](https://github.com/allyjweir/pg-trace-example/blob/master/src/exampleRouter.ts#L10) used pg via async/await (promises) and didn't have any issue with this.

Agree though that should go back and revisit this, testing all the different ways `.query()` can be called to be sure it works correctly. I will build out some more test cases.

lib/instrumentation/pg.js
allyjweir: Ran some tests and it turns out this is necessary! Agree on the better comment.

lib/instrumentation/pg.js
allyjweir: Have exhaustively tested all variations and they all work with current implementation.

| Endpoint                                                      | Status  | Fields Returned                                     |
|---------------------------------------------------------------|---------|-----------------------------------------------------|
| http GET localhost:8080/example/callback/text-only            | Success | query and rows_affected                             |
| http GET localhost:8080/example/callback/with-values          | Success | query, query_args and rows_affected                 |
| http GET localhost:8080/example/callback/with-config-obj      | Success | query, query_args and rows_affected                 |
| http GET localhost:8080/example/callback/prepared-stmt        | Success | query, query_args and rows_affected and query_name  |
| http GET localhost:8080/example/promises/text-only            | Success | query and rows_affected                             |
| http GET localhost:8080/example/promises/with-values          | Success | query, query_args and rows_affected                 |
| http GET localhost:8080/example/promises/with-config-obj      | Success | query, query_args and rows_affected                 |
| http GET localhost:8080/example/promises/prepared-stmt        | Success | query, query_args and rows_affected and query_name  |
| http GET localhost:8080/example/async-await/text-only         | Success | query and rows_affected                             |
| http GET localhost:8080/example/async-await/with-values       | Success | query, query_args and rows_affected                 |
| http GET localhost:8080/example/async-await/with-config-obj   | Success | query, query_args and rows_affected                 |
| http GET localhost:8080/example/async-await/prepared-stmt     | Success | query, query_args and rows_affected and query_name  |

lib/instrumentation/pg.js
toshok: 👍 later.  thanks for checking :)

lib/instrumentation/pg.js
toshok: for additional context for future features, I think the problem was entirely in my reading of the PR vs the api - the place where you're shimming is not the API I was looking at.

yay for exhaustive testing.

PR #158, opened 2019-08-08
update express instrumentation to correctly track `request.route` and `request.base_url`

tikotzky:
Currently it add `request.route` to the context before express matches the request to a route.
This PR moves `request.route` to the `boundFinisher` so that it will only be added after express has already matched the request to a route and set `request.route`.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -5,3 +5,4 @@ Christine Yen
 Erwin van der Koogh
 Ally Weir
 Ashish Bista
+Mordy Tikotzky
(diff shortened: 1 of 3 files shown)


tikotzky:
The same issue was happening with `request.base_url`.
It was setting it to the context as soon as the request came in, but express updated `req.baseUrl` as the request passes through the middleware.
In order for `request.route` to be set to the correct `req.baseUrl` it would need to be set to context after the request has already been handled by a route.

tikotzky:
Thanks.
Name added.

PR #162, opened 2019-08-21
Friendly error message if a trace doesn't exist when starting a span

carterbancroft:
Before, when starting a span outside of a trace it would throw `Cannot read property 'stack' of undefined` from the `context.stack.length` line.

This was tricky for me to figure out because in some unit tests much of our code would be run via an express endpoint call (which the beeline would instrument automatically) but in other cases where we're testing a particular unit outside of an express call it would fail. Until I properly understood context I couldn't grok why certain tests against code with spans would pass and others would fail.

A clearer error would hopefully make it more obvious what's happening.

I'm sort of guessing at a proper error message... if you like these changes but not the message let me know what might be better.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -6,3 +6,4 @@ Erwin van der Koogh
 Ally Weir
 Ashish Bista
 Mordy Tikotzky
+Carter Bancroft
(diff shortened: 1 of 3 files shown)


carterbancroft:
Hmm, I can't see why that build failed. The tests seem to pass locally and the linter seems happy.

toshok:
looks like a circleci billing thing (although I didn't think we had contributor limits on open source repos).  will investigate.

carterbancroft:
Ohh, I see, thanks. I saw that error but I thought it might be because I didn't have permissions to look at the build or something. Thought maybe it was hiding some other issue.

MikeGoldsmith:

PR #163, opened 2019-08-28
Adding support for setting user requested parent span id

praveengunupati:
Right now at lifion we are using existing trace and span Ids generated at our end instead of beeline generated Ids. As there is no way to tell beeline about user generated span Id, the spans are loading into honeycomb with out parent ids. As a result we are seeing lot missing parent spans and broken traces.
This PR provides support for adding custom  parent span Id.
If needed users can pass parentIdSource value while initialising beeline module .

### DIFF
--- a/lib/instrumentation/express.js
+++ b/lib/instrumentation/express.js
@@ -180,7 +180,7 @@ let instrumentExpress = function(express, opts = {}) {
       parentIdSource = opts.parentIdSource;
     } else {
       debug(
-        "parentIdSource option must either be an string"
+        "parentIdSource option must  be a string"
       );
     }
   }
--- a/lib/instrumentation/express.js
+++ b/lib/instrumentation/express.js
@@ -79,7 +79,11 @@ function wrapErrMiddleware(middleware, debugWrap) {
 }
 
 const getMagicMiddleware = ({ userContext, traceIdSource, parentIdSource, packageVersion }) => (req, res, next) => {
-  let traceContext = traceUtil.getTraceContext(traceIdSource, parentIdSource, req);
+  let traceContext = traceUtil.getTraceContext(traceIdSource, req);
+  let parentTraceId = traceUtil.getParentSourceId(parentIdSource, req);
+  if(parentTraceId){
+    traceContext["parentSpanId"] = parentTraceId;
+  }
   let trace = api.startTrace(
     {
       [schema.EVENT_TYPE]: "express",
@@ -176,11 +180,11 @@ let instrumentExpress = function(express, opts = {}) {
   }
 
   if (opts.parentIdSource) {
-    if (typeof opts.parentIdSource === "string") {
+    if (typeof opts.parentIdSource === "string" || typeof opts.traceIdSource === "function") {
       parentIdSource = opts.parentIdSource;
     } else {
       debug(
-        "parentIdSource option must  be a string"
+        "parentIdSource option must either be an string (the http header name) or a function returning the string request id"
       );
     }
   }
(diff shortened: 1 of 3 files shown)


lib/instrumentation/trace-util.js
```diff
@@ -26,7 +26,7 @@ exports.getTraceContext = (traceIdSource, req) => {
     let headers =
       typeof traceIdSource === "undefined"
         ? [api.TRACE_HTTP_HEADER, "X-Request-ID", "X-Amzn-Trace-Id"]
-        : [traceIdSource];
+        : [traceIdSource.split(";")[0]];
```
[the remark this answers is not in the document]
praveengunupati: Thanks @lizthegrey for the feedback. I have updated my PR with your suggestion.

lib/instrumentation/express.js
```diff
@@ -175,14 +175,24 @@ let instrumentExpress = function(express, opts = {}) {
     }
   }
 
+  if (opts.parentIdSource) {
+    if (typeof opts.parentIdSource === "string") {
+      parentIdSource = opts.parentIdSource;
+    } else {
+      debug(
+        "parentIdSource option must  be a string"
```
toshok: given that this is mirroring the traceIdSource option, it should probably take either a string or a function.

lib/instrumentation/trace-util.js
```diff
@@ -46,10 +46,14 @@ exports.getTraceContext = (traceIdSource, req) => {
       }
 
       default: {
-        return {
+        let traceContext = {
           traceId: value,
           source: `${header} http header`,
         };
+        if (typeof parentIdSource !== "undefined") {
+          traceContext["parentSpanId"] = req.headers[parentIdSource]
```
toshok: feels like this needs better integration here.  now `traceContext` holds two ids but only one source (the traceId source).

also, `parentIdSource` will only be used if `traceIdSource` is undefined or a string (nothing in the else case below.)

praveengunupati: @toshok I know you are looking for more streamlined code , but my intention here is to fix the problem that we have right now(we send traceIsSource as string and no TRACE_HTTP_HEADER). as you know this is hindering us getting the distributed view , can you please move this fix and will work on adding this new feature(custom parentSourceID) with another PR.

eanakashima:
Re-added to @lizthegrey's review queue to look at the changes she requested. Re-added to @toshok's review queue to answer the question in the line comments, about whether there's a fast route to unblocking this PR, if more work can be done in a follow-up.

@praveengunupati sorry about the build errors. I believe next time you push, the build should be working.

praveengunupati:
> Re-added to @lizthegrey's review queue to look at the changes she requested. Re-added to @toshok's review queue to answer the question in the line comments, about whether there's a fast route to unblocking this PR, if more work can be done in a follow-up.
> 
> @praveengunupati sorry about the build errors. I believe next time you push, the build should be working.

praveengunupati:
@eanakashima ,  @lizthegrey requested changes are completed and she moved the review to @toshok. I know this adding parentSourceId is new feature and they are lot more possibilities to cover, but I have added this feature in one of the case that is specific to lifion(ADP). I don't see any issue in merging the changes as it won't effect the existing features and create a follow up PR for the other changes.

praveengunupati:
@eanakashima @lizthegrey @toshok I have taken out the parentSourceID code from getTraceContext function and added a separate function to read parentSpanId if it presents in the input options and passing on to start trace. parentSourceId can now be passed as both string and function.

PR #168, opened 2019-10-12
[pg] we need to return the promise

toshok:
our `query` shim was executing the query properly if instrumentation was active, but it wasn't returning the promise back, leading to

```
TypeError: Cannot read property 'then' of undefined
```

This should fix things.


toshok:
if we'd written this in coffeescript we'd be home by now

eanakashima:
ahahaha i was just about to write something similar ☕️ 4eva

PR #175, opened 2019-11-21
PG instrumentation callbacks

### LINKED ISSUE #173: pg.Client#query instrumentation breaks pg-query-stream usage
Hi folks! Thanks for your hard work on this instrumentation!

I ran into a problem at work (at eaze) using [pg-query-stream](https://github.com/brianc/node-pg-query-stream) with `honeycomb-beeline`. This module (from brianc, the lead maintainer of `pg`) is intended to allow users to stream results back from postgres queries, vs. receiving them in a batch chunk. You can see [here in `pg`](https://github.com/brianc/node-postgres/blob/master/lib/client.js#L464-L469) that `pg.Client#query` is checking to see if the argument sent to the function has a `.submit` method, & if so it passes that item through. Similarly, `pg-query-stream` exposes that `submit()` method [here](https://github.com/brianc/node-pg-query-stream/blob/master/index.js#L22-L24).

Used in isolation, without `honeycomb-beeline`, this works fine. However, when we enable the instrumentation and call `postgresClient.query(pgQueryStream)`, the `pgQueryStream` is wrapped in a callback by [this instrumentation](https://github.com/honeycombio/beeline-nodejs/blob/master/lib/instrumentation/pg.js#L68), which turns the parameter into a `function`, hiding the `submit` method.

I think I have a workaround for the time being, but it'd be lovely if `honeycomb-beeline` could detect when arguments are passed with `.submit()` and wrap them as a stream instead of as a callback.

Thanks again for taking a look at this!

martin308:
Try to better handle the arguments being passed to the `query` function. It looks like having these three arguments have been pretty stable for many years. 

We need to decipher where in the arguments the callback lives. From reading the source, it looks like the options are these:

// query string, values, callback
// query string, callback
// Query like object
// Query like object, callback

Once we have deciphered where the correct callback lives we wrap it and pass it onto the wrapped `query` function.

This seems to work in my local manual testing. Looking into how to add some unit tests for this as well.

Resolves #173.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,13 @@ executors:
         default: "12"
     docker:
       - image: circleci/node:<< parameters.nodeversion >>
-
+        environment:
+          PGUSER: root
+          PGDATABASE: circle_test
+      - image: circleci/postgres:9-alpine-ram
+        environment:
+          POSTGRES_USER: root
+          POSTGRES_DB: circle_test
 jobs:
   test_libhoney:
     parameters:
--- a/lib/instrumentation/pg.js
+++ b/lib/instrumentation/pg.js
@@ -76,6 +76,10 @@ let instrumentPg = function(pg, opts = {}) {
               config.callback = wrapper(span, config.callback);
             } else if (typeof values === "function") {
               values = wrapper(span, values);
+            } else {
+              if (typeof config.on === "function") {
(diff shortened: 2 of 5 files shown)


martin308:
@toshok @chrisdickinson I'd love for you both to review this PR!

PR #180, opened 2020-02-19
[fastify] only wrap the request/reply hooks

toshok:
Add a whitelist for request/reply hooks (https://github.com/fastify/fastify/blob/master/docs/Hooks.md), and just pass the others through unshimmed (the shim code assumes we're dealing with those hooks currently.)

Also make things a bit more robust by not crashing if there are requests that somehow didn't start a trace.

### DIFF
--- a/lib/instrumentation/fastify.js
+++ b/lib/instrumentation/fastify.js
@@ -48,6 +48,19 @@ const instrumentFastify = function(fastify, opts = {}) {
   const trackedByRequest = new Map();
   const finishersByRequest = new Map();
 
+  // we only wrap the hooks that deal with requests.  we should look into wrapping the non request/reply hooks at
+  // some point.
+  const wrappedHookNames = [
+    "onRequest",
+    "preParsing",
+    "preValidation",
+    "preHandler",
+    "preSerialization",
+    "onError",
+    "onSend",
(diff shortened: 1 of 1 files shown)


lib/instrumentation/fastify.js
```diff
@@ -127,16 +140,18 @@ const instrumentFastify = function(fastify, opts = {}) {
     // others will show up as spans in the trace.
     shimmer.wrap(app, "addHook", function instrumentAddHook(original) {
       return function(hookName, hookFn) {
-        if (hookName == "onResponse") {
-          // we need to skip onResponse hooks, because we will have already ended the
-          // trace.  we need a way to manipulate onResponse hooks
```
martin308: is this comment not valid anymore? It looks like we will now be using the `onResponse` hook?

toshok: oh derp.  I need to remove that from the whitelist.  thanks :)

toshok: to be clear - the robust-change will make it so we can shim `onResponse`, but the comment is still valid, in that the trace will have already been sent.  It'd be nice to instrument `onResponse`, so left the comment in (also commented it out of the white list since there's no reason to shim it currently.)

Issue #141, opened 2019-05-29
Parse Root= from X-Amzn-Trace-Id when ALB header is used


nathanleclaire:
I think all of X-Amzn-Trace-Id is used for `trace.trace_id` which is a bit much

camerondunn-ns8:
@nathanleclaire Is this in reference to the parsing of a downstream 'X-Amzn-Trace-Id' header only capturing "Root" from an original ALB header?

For example, I have a situation where calls flow as:
ALB -> Service1 -> Service2. Both Service1 & Service2 are using beeline-nodejs

The behavior I see is:
1) ALB sets headers like: "X-Amzn-Trace-Id: Root=<someID>"
2) Service1's beeline parses "X-Amzn-Trace-Id" since no other trace header is present. Sets the 'trace_id' context to be "Root=<someID>".
3) When an outbound call from Service1 is made, the "X-Honeycomb-Trace" header is set to something like "1;trace_id=Root=<someID>,parent_id=<AnotherID>,dataset=<OutDataset?,context=<Base64Stuff>="
4) Service2 parses "X-Honeycomb-Trace" by splitting on "=" and reads the "trace_id" as "Root", not "Root=<someID>".

In the Honeycomb UI I can see a trace page where ALL requests to Service2 are considered part of one trace. I changed our traceID definition for our dataset to something else because I was afraid this wouldn't scale and would break the UI.

Is that the issue you're referring to? Is there a plan to fix this, or would you accept a pull request?

nathanleclaire:
Haven't had a chance to review your comment in detail but we definitely welcome PRs!

camerondunn-ns8:
@nathanleclaire PR can be found [here](https://github.com/honeycombio/beeline-nodejs/pull/182).

paulosman:
This should be fixed in #187

PR #182, opened 2020-02-28
Pull prefix & "=" out of x-amzn-trace-id

camerondunn-ns8:
Leaving these in enables a problematic situation where if the
full x-amzn-trace-id is set as the trace id, then a downstream
service will only pull out the prefix of the original trace id
(e.g. "Root") and ALL requests will be considered part of a common
trace.

## Background
I believe this addresses this issue: https://github.com/honeycombio/beeline-nodejs/issues/141

## Verification
Unit tests validate specific behavior. If there are concerns I can try and get these changes behind an ELB.

### DIFF
--- a/lib/instrumentation/trace-util.js
+++ b/lib/instrumentation/trace-util.js
@@ -21,11 +21,13 @@ const getValueFromHeaders = (requestHeaders, headers) => {
   return undefined;
 };
 
+const X_AMZN_TRACE_ID_HEADER = "X-Amzn-Trace-Id";
+
 exports.getTraceContext = (traceIdSource, req) => {
   if (typeof traceIdSource === "undefined" || typeof traceIdSource === "string") {
     let headers =
       typeof traceIdSource === "undefined"
-        ? [api.TRACE_HTTP_HEADER, "X-Request-ID", "X-Amzn-Trace-Id"]
+        ? [api.TRACE_HTTP_HEADER, "X-Request-ID", X_AMZN_TRACE_ID_HEADER]
         : [traceIdSource];
     let valueAndHeader = getValueFromHeaders(req.headers, headers);
 
@@ -43,6 +45,14 @@ exports.getTraceContext = (traceIdSource, req) => {
         return Object.assign({}, parsed, {
           source: `${header} http header`,
         });
+      }
+
+      case X_AMZN_TRACE_ID_HEADER: {
+        return {
+          // Don't pass along "Root=" as it can break parsing further down the chain.
+          traceId: value.split("=")[1],
+          source: `${header} http header`,
+        };
       }
 
       default: {
@@ -66,7 +76,7 @@ exports.getParentSourceId = (parentIdSource, req) => {
   } else if (typeof traceIdSource === "function") {
     return parentIdSource(req);
   }
-};  
+};
 
 exports.getUserContext = (userContext, req) => {
   if (!userContext) {
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -6,3 +6,4 @@ Erwin van der Koogh
 Ally Weir
 Ashish Bista
 Mordy Tikotzky
+Cameron Dunn
--- a/lib/instrumentation/express.test.js
+++ b/lib/instrumentation/express.test.js
@@ -232,7 +232,7 @@ describe("request id from http headers", () => {
       .expect(200, () => {
         expect(api._apiForTesting().sentEvents.length).toBe(1);
         let ev = api._apiForTesting().sentEvents[0];
-        expect(ev[schema.TRACE_ID]).toBe("Root=1-67891233-abcdef012345678912345678");
+        expect(ev[schema.TRACE_ID]).toBe("1-67891233-abcdef012345678912345678");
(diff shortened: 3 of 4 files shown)


maplebed:
I think this logic needs a little more complexity. The AWS docs at https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html describe the syntax for the tracing header (`version-timestamp-traceid`) and we should follow that spec. It also describes both the `Root` and `Self` portions of the header; we'll need to parse those in order to understand requests that flow through multiple services. 

There are two spots that have done this implementation (albeit in go rather than js) that we can use for reference: https://github.com/honeycombio/honeyaws/pull/55/files and https://github.com/honeycombio/beeline-go/pull/88/files#diff-d3871234ed82727e61f5e5ee2add1ed8R125. I haven't looked at them closely enough to understand the differences (though there are some).

camerondunn-ns8:
> I think this logic needs a little more complexity. The AWS docs at https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html describe the syntax for the tracing header (`version-timestamp-traceid`) and we should follow that spec. It also describes both the `Root` and `Self` portions of the header; we'll need to parse those in order to understand requests that flow through multiple services.

Sure, so the changes which I think are being asked for:
* Both implementations use only the traceId component out of the version-timestamp-traceid string.
  * I just want to point out this additional parsing isn't "needed" for compatibility. This just would bring JS in line w/ the other implementations. That is fine by me, but it might be good to document this is if it's assumed to be universal. Users of this library will also see a more significant change to the values of their trace IDs.
* Graceful handling of chained x-amzn-trace-id headers.
  * E.g. with these changes the header `X-Amzn-Trace-Id: Self=1-67891234-12456789abcdef012345678;Root=1-67891233-abcdef012345678912345678` would result in the traceId being: `1-67891234-12456789abcdef012345678;Root` -- technically compatible but probably not what we want.
  * In the 'beeline-go' it supports these chained traces, but you'd only see these chains if the upstream system is actually setting the 'X-Amzn-Trace-Id' (and not the honeycomb trace header).
* Check beginning (Root) and end (last Self) traces in chain in order to get both ParentID and TraceID.
  * Rest of my comment are in regard to this.

Below is just a description of behavior and a question -- at first I thought this didn't seem correct, but now I think it's Ok & just a little weird. When the beeline sees a header like:
``X-Amzn-Trace-Id: Self=1-67891234-12456789abcdef012345678;Root=1-67891233-abcdef012345678912345678`

it will, I believe, create a propagation context of
```
{
  TraceID: "12456789abcdef012345678",
  ParentId: "abcdef012345678912345678"
  // SpanId is unset (which is correct I think), but the comments refer to the spanID several times which makes the intended behavior confusing. Maybe this was originally looking at the honeyaws code as reference originally?
}
```

I want to clarify this behavior in the situation where we have
ALB -> Service1 (not enabled with Honeycomb, does pass along 'X-Amzn-Trace-Id' untouched) -> ALB -> Service2 (using beeline-go).

The first event sent to honeycomb (from Service2) will include a traceID, a parent span, and the current span ID (generated outside of the context we're looking at). Does honeycomb handle references to parentID (parent span) with no corresponding actual span (e.g. nothing includes a spanID w/ the corresponding ID)? The other odd thing about this is that if you add additional non-honeycomb services in between Service1 and Service2, the Honeycomb UI would suggest that the calls from Service1 to Service2 were direct (well just the ELB between).

Hope that all makes some sense! I can make all of these changes, but want to verify intent beforehand. I'm not a go expert and the beeline-go PR is still open, and the honeyaws library has different intentions since its parsing ALB logs and is describing the ALB as an actual metric/piece of work.

lib/instrumentation/trace-util.js
```diff
@@ -45,6 +47,14 @@ exports.getTraceContext = (traceIdSource, req) => {
         });
       }
 
+      case X_AMZN_TRACE_ID_HEADER: {
+        return {
+          // Don't pass along "Root=" as it can break parsing further down the chain.
+          traceId: value.split("=")[1],
```
paulosman: small thing, but can we replace `[1]` with something like `pop()`? I'm worried about the case where malformed input could cause this to be `undefined`.

paulosman:
@camerondunn-ns8 This is great, thank you. Because it deals with interactions between a Honeycomb beeline and a 3rd party (AWS) we want to make sure that whatever we do, we do it consistently across all beelines and integrations. We've got a design doc going weighing various approaches and will incorporate this PR (possibly with changes) into the solution once we're ready to roll it out. I'm going to close this PR for now but this is on our roadmap for beelines and will be fixed as soon as we decide on an implementation that can be shared across integrations. I'll make sure to @-mention you when the implementation does get submitted to make sure it fixes your specific issue.

camerondunn-ns8:
@paulosman do you have a timeline for your roadmap?

paulosman:
@camerondunn-ns8 within the next couple of weeks seems reasonable.

paulosman:
@camerondunn-ns8 we just released [v2.1.0](https://github.com/honeycombio/beeline-nodejs/releases/tag/v2.1.0). Can you verify that this fix works for you?

PR #183, opened 2020-03-03
Remove entries from previously unbounded maps

camerondunn-ns8:
* The request is safe to delete from finishersByRequest because
it is accessed only once, on the previous line, so long as there
is only one 'onResponse' per request.
* The request is safe to delete from trackedByRequest because
onResponse is the last available hook in the lifecycle.

## Verification
I've run this locally and validated using a debugger that these maps don't grow anymore. However I'm not confident enough in my understanding of Fastify hooks/lifecycle to know if there are cases where onResponse is called multiple times for a request, out of order, etc.

It might be worth just adding in some error catching and logging if the deletes fail or entries are missing.


toshok:
🎉

PR #184, opened 2020-03-06
clean up a bunch of span/context related hacks and bugs

### LINKED ISSUE #149: Documentation for beeline.addContext is wrong
https://docs.honeycomb.io/getting-data-in/javascript/beeline-nodejs/

> Note that beeline.addContext only adds fields to the current span. In order to add fields to all downstream spans, use beeline.customContext.add instead:

From reading the code and experimenting with the SDK it seems that `beeline.addContext()` actually adds the context to every active/subsequent span in the trace. This is fine, and useful, but it seems there's no way to add context to the currently active span without keeping a reference to it that you pass around and manually add entries to?

It seems this behaviour was changed in: https://github.com/honeycombio/beeline-nodejs/pull/54

https://honeycombpollinators.slack.com/archives/C4T3A32CX/p1563529998355500

### LINKED ISSUE #156: `addContext` and `customContext.add` are similar but accept different args
`addContext()` accepts a map of arguments, but `customContext.add()` only accepts a single key/value, which means you need to invoke it for every piece of metadata you want to add to the trace. It'd be easier if you could pass a map or a single key/value, or make sure the caller _always_  passes a map.

toshok:
A few things changed here - one thing in a backward incompatible way _but_ in a way consistent with our docs.

1. Introduce a `Span` class, and change the api implementations to use instances of `Span` instead of regular JS objects in their `context.stack`.
2. `Span` exposes an `addContext` method that acts similar to the go beeline's -- no prefixing of `app.` when using it.  Change all instrumentation to call this method directly, since we always have a span there.
3. `Span` also has fields for `startTime`/`startTimeHR` so all the handling of those fields can be cleaned up - no more `delete`ing them from the payload when we send (and in tests!  ugh.)
3. Add a toplevel `addTraceContext` api method that acts similar to the go beeline's -- prefix with `app.` and also propagates to outbound http/https.
4. The backward incompatible change alluded to above: `api.addContext` no longer results in that context being propagated to outbound calls.  This is the documented behavior, so I don't feel as bad about making this change.

Now for the good stuff.  This PR also marks some things as deprecated or to-be-changed with the next major bump:

1. `api.customContext.{add, remove}` goes away entirely.  `.add` is the same as `api.addTraceContext`.  there is no replacement for `.remove`
2. similarly, `api.removeContext` goes away without replacement.
3. `api.addContext` will prefix with `app.`.  This mirrors the go beeline - if you have direct reference to a span you can call `span.addContext` and it won't prepend `app.`.  If you call the toplevel `addContext`, it will.

Added some more tests that exercise the various context adding methods so we'll fail those if we have regressions again in any of the documented behavior.

Fixes #149 
Fixes #156 
maybe others...

### DIFF
--- a/lib/api/index.test.js
+++ b/lib/api/index.test.js
@@ -480,8 +480,6 @@ test("distributed tracing linkage is correct", () => {
 
   let rootContext = tracker.getTracked();
 
-  console.log(rootSpan.payload, rootContext.traceContext);
-
   let marshaledContext = propagation.marshalTraceContext(rootContext);
 
   // this little bit of code simulates a downstream service.
--- a/lib/api/index.test.js
+++ b/lib/api/index.test.js
@@ -493,7 +493,7 @@ test("distributed tracing linkage is correct", () => {
     ctx.parentSpanId
   );
   const subContext = tracker.getTracked();
-  subContext.traceContext = ctx.traceContext;
+  subContext.traceContext = ctx.customContext;
 
   let subserviceSubspan = api.startSpan({
     [schema.TRACE_SPAN_NAME]: "subspan",
--- a/lib/api/index.test.js
+++ b/lib/api/index.test.js
@@ -1,6 +1,6 @@
 /* eslint-env node, jest */
 const path = require("path"),
-  event = require("."),
+  api = require("."),
   schema = require("../schema"),
   tracker = require("../async_tracker"),
   propagation = require("../propagation"),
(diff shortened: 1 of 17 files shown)


lib/propagation.js
```diff
@@ -89,15 +89,15 @@ function unmarshalTraceContextv1(payload) {
...
-      customContext = JSON.parse(Buffer.from(contextb64, "base64").toString("ascii"));
+      traceContext = JSON.parse(Buffer.from(contextb64, "base64").toString("ascii"));
     } catch (e) {
       debug("couldn't decode context", e);
       return;
     }
   }
 
-  return { traceId, parentSpanId, customContext, dataset };
+  return { traceId, parentSpanId, traceContext, dataset };
```
martin308: is this key name change a breaking change?

lib/api/index.test.js
```diff
@@ -463,38 +465,48 @@ describe("custom context", () => {
...
+
+  // add some additional context to the root span
+  rootSpan.addContext({ directContext: "direct" });
+  api.addContext({ localContext: "local" });
+  api.customContext.add({ customContext: "custom" });
+  api.addTraceContext({ traceCustomContext: "trace-custom" });
 
   let rootContext = tracker.getTracked();
 
+  console.log(rootSpan.payload, rootContext.traceContext);
```
martin308: should this still be here?

toshok: nope! :)

lib/propagation.js
toshok: hm, yeah.  the documentation is actually lacking on this return value (it only mentions traceId and parentSpanId), but the object isn't meant to be an opaque thing that gets passed back into the API (really it should be..)  will change it back to customContext and add a note to rename it next major bump.

PR #187, opened 2020-03-19
parse AWS trace header

toshok:
We care about Root/Parent clauses currently, and map those to traceId/parentSpanId.

This code matches the behavior of code aws uses in their sdks.

We'll need to figure out how propagation works in the aws world (on outbound requests) so we can add our custom stuff there as well, which will in turn cause more changes to this code.

algorithm for dealing with the header is basically:

1. split aws header by `;`
2. walk the list of strings from #1, splitting each string by `=` into [key, value].
    2a. if we see key="Root", value becomes traceId.
    2b. if we see key="Parent", value becomes parentSpanId.
3. If we didn't see traceId, bail.
4. If we didn't see parentSpanId, fall back to traceId.

### DIFF
--- a/lib/instrumentation/trace-util.js
+++ b/lib/instrumentation/trace-util.js
@@ -65,6 +65,10 @@ exports.getTraceContext = (traceIdSource, req) => {
           return {};
         }
 
+if (!parentSpanId) {
+  parentSpanId = traceId;
+}
+
         return {
           traceId,
           parentSpanId,
--- a/lib/instrumentation/trace-util.js
+++ b/lib/instrumentation/trace-util.js
@@ -21,11 +21,13 @@ const getValueFromHeaders = (requestHeaders, headers) => {
   return undefined;
 };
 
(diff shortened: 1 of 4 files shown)


lib/instrumentation/trace-util.js
```diff
@@ -45,6 +47,31 @@ exports.getTraceContext = (traceIdSource, req) => {
...
+          } else if (name === "Parent") {
+            parentSpanId = value;
+          }
+        }
+
+        if (!traceId) {
+          // if we didn't even get a 'Root=' clause, bail.
+          return {};
+        }
+
```
toshok: one thing here.  @maplebed beeline-go PoC did something like this here:

```suggestion
if (!parentId) {
  parentId = traceId;
}
```

should 100% do that here.

paulosman: Yep - I like this and think we should do it everywhere. Seems to make sense.

PR #189, opened 2020-03-21
Attempt to fix pg stream support for v3 of pg

martin308:
This should be a PR into #188 but as it is a fork I can't do that. If it works maybe you can cherry-pick the commit @toshok?


martin308:
Incorrect set of commits to put this on

PR #193, opened 2020-04-02
Make non-Honeycomb header trace parsing opt-in.

paulosman:
Before this change, we've been automatically parsing incoming Amazon
trace ID headers, and recently we started using the extracted values to
set trace.trace_id and trace.parent_id. This is potentially surprising
to users who are not using honeyaws to parse their ALB or ELB logs and
send the data to the same dataset, in that they'd get traces with
missing root spans.

This change requires that users of the node beeline specify the trace
source in the beeline initialization, e.g.:

```
    writeKey:    "foo",
    dataset:     "bar",
    serviceName: "foo-service",
    express: {
        traceIdSource: 'X-Amzn-Trace-Id',
    },
})
```
Note this is a breaking change so will have to be included in a major
release of the beeline.

### DIFF
--- a/lib/instrumentation/express.test.js
+++ b/lib/instrumentation/express.test.js
@@ -129,7 +129,7 @@ describe("userContext as function", () => {
   });
 });
 
-describe("req proprties should only be read after a request matches a route", () => {
+describe("req properties should only be read after a request matches a route", () => {
   const express = instrumentExpress(require("express"), {});
 
   let server;
@@ -185,8 +185,8 @@ describe("req proprties should only be read after a request matches a route", ()
   });
 });
 
-describe("request id from http headers", () => {
-  const express = instrumentExpress(require("express"), {});
+describe("request id from x-request-id http header", () => {
+  const express = instrumentExpress(require("express"), { traceIdSource: "X-Request-ID" });
 
   let server;
   function initializeTestServer() {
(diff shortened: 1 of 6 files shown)


paulosman:
Addressing test failures...

lib/propagation.js
```diff
@@ -4,7 +4,9 @@ const path = require("path"),
   debug = require("debug")(`${pkg.name}:propagation`),
   schema = require("./schema");
 
-exports.TRACE_HTTP_HEADER = "X-Honeycomb-Trace";
+exports.TRACE_HTTP_HEADER  = "X-Honeycomb-Trace";
+exports.AMAZON_HTTP_HEADER = "X-Amzn-Trace-Id";
+exports.REQUEST_ID_HEADER  = "X-Request-ID";
```
toshok, line 9: bikeshedding here, but wdyt about adding TRACE to amazon, and HTTP to request-id?

```
exports.AMAZON_TRACE_HTTP_HEADER = "X-Amzn-Trace-Id";
exports.REQUEST_ID_HTTP_HEADER  = "X-Request-ID";
```

wordier to be sure.

lib/instrumentation/express.test.js
```diff
@@ -185,8 +185,8 @@ describe("req proprties should only be read after a request matches a route", ()
   });
 });
 
-describe("request id from http headers", () => {
-  const express = instrumentExpress(require("express"), {});
+describe("request id from x-request-id http header", () => {
+  const express = instrumentExpress(require("express"), { traceIdSource: "X-Request-ID" });
```
toshok, line 189: should this use the api export for the trace header?

lib/instrumentation/express.test.js
```diff
@@ -220,6 +220,30 @@ describe("request id from http headers", () => {
         done();
       });
   });
+});
+
+describe("Trace ids from X-Amzn-trace-id header", () => {
+  const express = instrumentExpress(require("express"), { traceIdSource: "X-Amzn-Trace-Id" });
```
toshok, line 226: same here.

lib/instrumentation/express.test.js
```diff
@@ -233,6 +257,30 @@ describe("request id from http headers", () => {
         done();
       });
   });
+});
+
+describe("Trace ids from X-Request-ID and X-Amzn-trace-id headers", () => {
+  const express = instrumentExpress(require("express"), { traceIdSource: "X-Request-ID" });
```
toshok, line 263: here too.

lib/propagation.js
paulosman: I was wondering the same thing actually. Thanks for confirming. Will update :)

lib/instrumentation/express.test.js
paulosman: Yes, thank you - will update both

PR #194, opened 2020-04-02
Make non-Honeycomb header trace parsing opt-in.

paulosman:
Make header constants consistent. Add tests for setting `traceIdSource` explicitly in the express of fastify initialization code. Users who do not want automatic parsing of incoming Amazon trace ID headers or Request IDs should now specify a `traceIdSource` in their initialization code to prevent missing root spans.

```
    writeKey:    "foo",
    dataset:     "bar",
    serviceName: "foo-service",
    express: {
        traceIdSource: 'X-Amzn-Trace-Id',
    },
})
```

In this next major release, we'll make this behavior the default (only Honeycomb headers will be parsed unless another header format is explicitly specified).

### DIFF
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -50,6 +50,8 @@ module.exports = {
   },
 
   TRACE_HTTP_HEADER: propagation.TRACE_HTTP_HEADER,
+  AMAZON_HTTP_HEADER: propagation.AMAZON_HTTP_HEADER,
+  REQUEST_ID_HEADER: propagation.REQUEST_ID_HEADER,
   marshalTraceContext: propagation.marshalTraceContext,
   unmarshalTraceContext: propagation.unmarshalTraceContext,
 
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -50,7 +50,7 @@ module.exports = {
   },
 
   TRACE_HTTP_HEADER: propagation.TRACE_HTTP_HEADER,
-  AMAZON_HTTP_HEADER: propagation.AMAZON_HTTP_HEADER,
+  AMAZON_TRACE_HTTP_HEADER: propagation.AMAZON_TRACE_HTTP_HEADER,
   REQUEST_ID_HEADER: propagation.REQUEST_ID_HEADER,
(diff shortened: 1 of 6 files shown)


lib/api/index.js
```diff
@@ -50,6 +50,8 @@ module.exports = {
   },
 
   TRACE_HTTP_HEADER: propagation.TRACE_HTTP_HEADER,
+  AMAZON_TRACE_HTTP_HEADER: propagation.AMAZON_TRACE_HTTP_HEADER,
+  REQUEST_ID_HEADER: propagation.REQUEST_ID_HEADER,
```
toshok, line 54: think this one is still missing an `HTTP`?

```suggestion
  REQUEST_ID_HTTP_HEADER: propagation.REQUEST_ID_HTTP_HEADER,
```

lib/instrumentation/trace-util.js
```diff
@@ -21,14 +21,9 @@ const getValueFromHeaders = (requestHeaders, headers) => {
...
 
-const X_AMZN_TRACE_ID_HEADER = "X-Amzn-Trace-Id";
-
 exports.getTraceContext = (traceIdSource, req) => {
   if (typeof traceIdSource === "undefined" || typeof traceIdSource === "string") {
-    let headers =
-      typeof traceIdSource === "undefined"
-        ? [api.TRACE_HTTP_HEADER, "X-Request-ID", X_AMZN_TRACE_ID_HEADER]
-        : [traceIdSource];
+    let headers = typeof traceIdSource === "undefined" ? [api.TRACE_HTTP_HEADER] : [traceIdSource];
```
toshok, line 26: hm, could we leave this line the way it is (so can push a minor release) and detect if we autoconfigured against request-id or amazon trace header?  That way we can warn users that they'll need to be explicit about that in the next major release.  The warning might be enough of a cue for users to fix things so they won't see missing root spans.

paulosman: Yeah, I like that. Users can set `traceIdSource` to `X-Honeycomb-Trace` in the beeline initialization for example to prevent amazon trace headers from being parsed. I'll update.

PR #197, opened 2020-04-16
Fix access to TRACE_SPAN_ID as it now is part of the Span payload

### LINKED ISSUE #196: startAsyncSpan fails to capture parent span
After updating `beeline-nodejs` to 2.1.0, we started getting broken traces in HC. All spans are missing a parent span.

grafik

When debugging `startAsyncSpan`, I notice that `parentId` is undefined. Line https://github.com/honeycombio/beeline-nodejs/blob/efc06456001a3f972ef86b6da8ba81ee8e33481a/lib/api/libhoney.js#L151 is trying to fetch `parentId` from the last stack item, but this item is of type `Span` and contains the required property `trace.span_id` inside a property `payload`: https://github.com/honeycombio/beeline-nodejs/blob/efc06456001a3f972ef86b6da8ba81ee8e33481a/lib/api/span.js#L7 . So it can't find it, which leaves all spans started using `startAsyncSpan` without a parent span.

IMO this break all instances of `startAsyncSpan`, because they all fail to capture the parent span.

jugglingthebits:
This PR updates `startAsyncSpan` to fetch the parent span id from the span payload property.

Fixes #196

### DIFF
--- a/lib/api/libhoney.js
+++ b/lib/api/libhoney.js
@@ -148,7 +148,7 @@ module.exports = class LibhoneyEventAPI {
       return;
     }
     if (parentContext.stack.length > 0) {
-      parentId = parentContext.stack[parentContext.stack.length - 1][schema.TRACE_SPAN_ID];
+      parentId = parentContext.stack[parentContext.stack.length - 1].payload[schema.TRACE_SPAN_ID];
     }
     if (!parentId) {
       parentId = parentContext.parentId;


toshok:
awesome.  thanks - this pointed out there's some missing test coverage too.  going to add a test here before merging.

toshok:
actually no I'm not - pr remotes are hard :)

PR #200, opened 2020-05-12
bump libhoney to 2.1.0

toshok:
let's keep libhoney dep up to date so we're dogfooding.


toshok:
I'm guessing it's due to node 12 being used (was using 10 before).  will add a volta section/.node-version to the beeline in another PR so we're all at least on the same footing.  given https://github.com/honeycombio/libhoney-js/pull/70 is coming, might delay this and bump it again to 2.2.0 (or close and open a new one).

toshok:
closing this one since it's going to be superseded by a newer libhoney anyway

PR #206, opened 2020-06-04
Add common request headers to auto-instrumented fields.

paulosman:
The ruby beeline instruments Accept, Content-Type and User Agent. In addition, we want X-Forwarded-For and similar headers for cases when a service is deployed behind a proxy or load balancer and remote_addr might be obscured.

### DIFF
--- a/lib/instrumentation/express.js
+++ b/lib/instrumentation/express.js
@@ -88,24 +88,33 @@ const getMagicMiddleware = ({ userContext, traceIdSource, parentIdSource, packag
   if (parentTraceId) {
     traceContext.parentSpanId = parentTraceId;
   }
+  let fields = {
+    [schema.EVENT_TYPE]: "express",
+    [schema.PACKAGE_VERSION]: packageVersion,
+    [schema.TRACE_SPAN_NAME]: "request",
+    [schema.TRACE_ID_SOURCE]: traceContext.source,
+    "request.host": req.hostname,
+    "request.original_url": req.originalUrl,
+    "request.remote_addr": req.ip,
+    "request.secure": req.secure,
+    "request.method": req.method,
+    "request.scheme": req.protocol,
+    "request.path": req.path,
+    "request.query": req.query,
+    "request.http_version": `HTTP/${req.httpVersion}`,
+    "request.fresh": req.fresh,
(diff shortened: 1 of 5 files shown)


lib/instrumentation/express.js
```diff
@@ -88,24 +88,42 @@ const getMagicMiddleware = ({ userContext, traceIdSource, parentIdSource, packag
...
+  };
+
+  let header_names = {
+    "X-Forwarded-For": "request.header.x_forwarded_for",
+    "X-Forwarded-Proto": "request.header.x_forwarded_proto",
+    "X-Forwarded-Port": "request.header.x_forwarded_port",
+    "User-Agent": "request.header.user_agent",
+    "Content-Type": "request.header.content_type",
+    Accept: "request.header.accept",
+  };
```
toshok, line 116: couple of nits here:

1. camelCase, not snake.
2. I'd suggest pulling this out of this method and marking it `const`.
3. actually I'd pull it out even further to a shared location (a new file, probably), since it's duplicated in fastify.js (the code in fastify can `.toLowerCase()` the keys at lookup time.)

lib/instrumentation/fastify.js
```diff
@@ -75,20 +75,37 @@ const instrumentFastify = function(fastify, opts = {}) {
...
+        "request.method": request.req.method,
+        "request.route": request.route ? request.route.path : undefined,
+        "request.query": request.query,
+        "request.http_version": `HTTP/${request.req.httpVersion}`,
+      };
+
+      let header_names = {
+        "x-forwarded-for": "request.header.x_forwarded_for",
+        "x-Forwarded-proto": "request.header.x_forwarded_proto",
+        "x-Forwarded-port": "request.header.x_forwarded_port",
```
toshok, line 95: these keys aren't fully lowercased.  something `.toLowerCase()` would guard against :)

lib/instrumentation/express.js
paulosman: Updated!

lib/instrumentation/fastify.js
paulosman: That's not a nit, that's a straight-up bug :) Yeah, running `toLowerCase()` in the caller to the shared function fixes it.

PR #210, opened 2020-06-15
http(s) instrumentation should output urls with path/pathname/query correctly

numso:
If you call `axios.get("https://google.com/search?q=honeycomb.io")` then honeycomb will record it as `https://google.com/search%3Fq=honeycomb.io?q=honeycomb.io`.

axios parses the url and calls http.get with `{ path, pathname, search }`. The http/https instrumentation should respect the incoming pathname if it exists.


toshok:
let me take a crack at fixing the conflicts

vreynolds (approved):
Conflicts resolved. Thanks @numso !

PR #213, opened 2020-06-16
[http] make .get/.request work for both 2 (all node versions) and 3 (node 10.9.0+) arg versions

### LINKED ISSUE #204: http.request patch is incorrect since Node.js v10.9.0
Node.js v10.9.0 added a function signature to [http.request](https://nodejs.org/docs/latest-v12.x/api/http.html#http_http_request_options_callback) which accepts three arguments: `http.request(url[, options][, callback])`

This broke the patch in [http](https://github.com/honeycombio/beeline-nodejs/blob/master/lib/instrumentation/http.js#L12) and [https](https://github.com/honeycombio/beeline-nodejs/blob/master/lib/instrumentation/https.js#L13).

It also has the unfortunate side-effect of breaking other modules which make use of the new signature, in our case, [got](https://github.com/sindresorhus/got/issues/951).

The issue should be trivially reproducible: make a request with the latest version of got in an app where beeline-nodejs tracing is enabled.

toshok:
Node 10.9.0 added a 3 arg form for .get/.request in `http` and `https`, so we now have 2 forms to support:

```
http.request(options[, callback])
http.request(url[, options][, callback])
```

Instead of including a semver check in the instrumentation, handle the 3-arg form everywhere (this may or may not be the right approach, but it's definitely easier to write.)

Fixes #204

### DIFF
--- a/lib/instrumentation/http.js
+++ b/lib/instrumentation/http.js
@@ -8,36 +8,50 @@ let instrumentHTTP = (http, opts = {}) => {
   shimmer.wrap(http, "get", function(_original) {
     // we have to replace http.get since it references request through
     // a closure (so we can't replace the value it uses..)
-    return function(options, cb) {
-      var req = http.request(options, cb);
+    return function(_url, _options, _cb) {
+      let req = http.request.apply(this, Array.from(arguments));
       req.end();
       return req;
     };
   });
 
   shimmer.wrap(http, "request", function(original) {
-    return function(options, cb) {
+    // node 10+ supports a three arg form.  Instead of writing both implementations,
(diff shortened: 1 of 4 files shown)


toshok:
finally got this dusted off enough to find the error in the `https` instrumentation.  it was dropping the first argument from the `original.apply` call.

Fixed that call, and added tests for both http and https making requests without an active trace.  The https test failed before that last `instrumentation/https.js` change.

toshok:
sigh

image

toshok:
okay, should be good to go now.

Zertz:
Thanks to everyone involved, this will allow us to re-enable Honeycomb 😻

PR #215, opened 2020-06-30
Add null guard to addContext mock implementation

kenmclennan:
This PR brings the `addContext` implementation in `MockEventAPI` into
line with `LibhoneyEventAPI`.

The implementation of `addContext` in `LibhoneyEventAPI` checks that
a context exists before attempt to modify the context's stack.

The `MockEventAPI` implementation does not have this check. This can
cause tests to fail in services where the mock implementation is used
in test environments.

### DIFF
--- a/lib/api/mock.js
+++ b/lib/api/mock.js
@@ -83,6 +83,9 @@ module.exports = class MockEventAPI {
 
   addContext(map) {
     let context = tracker.getTracked();
+    if (!context || context.stack.length === 0) {
+      return;
+    }
     Object.assign(context.stack[0], map);
   }
 


kenmclennan:
@toshok I haven't added any tests here as the MockEventAPI is currently untested. This is something I can add if you like in a followup PR. Just let me know.

toshok:
nah, no need - I think leaving it untested is status-quo for the moment and I'll add task to make sure we write a test that exercises the API in general, and run it against both implementations.

toshok:
🎉

kenmclennan:
Awesome, thanks!

PR #216, opened 2020-07-01
switch UUID span and trace ids to W3C

katiebayes:
The node version of this: https://github.com/honeycombio/beeline-go/pull/110

We're switching over to compatible-with-W3C trace and span ids to smooth out support of OTel interop. Using pretty much the exact mechanism as open-telemetry.

For now I'm sticking this in propagation where it seemed to make the most sense, but propagation.js will be refactored in a subsequent PR.

### DIFF
--- a/lib/propagation.js
+++ b/lib/propagation.js
@@ -13,6 +13,9 @@ const VERSION = "1";
 exports.VERSION = VERSION;
 
 // id generation
+exports.generateSpanId = generateSpanId;
+exports.generateTraceId = generateTraceId;
+
 const SPAN_ID_BYTES = 8;
 const TRACE_ID_BYTES = 16;
 
@@ -23,9 +26,6 @@ function generateTraceId() {
 function generateSpanId() {
   return crypto.randomBytes(SPAN_ID_BYTES).toString("hex");
 }
-
-exports.generateSpanId = generateSpanId;
-exports.generateTraceId = generateTraceId;
 
 // assumes a header of the form:
 
--- a/lib/propagation.js
+++ b/lib/propagation.js
@@ -17,11 +17,11 @@ const SPAN_ID_BYTES = 8;
 const TRACE_ID_BYTES = 16;
 
 function generateTraceId() {
-  crypto.randomBytes(TRACE_ID_BYTES).toString("hex");
+  return crypto.randomBytes(TRACE_ID_BYTES).toString("hex");
 }
 
 function generateSpanId() {
-  crypto.randomBytes(SPAN_ID_BYTES).toString("hex");
+  return crypto.randomBytes(SPAN_ID_BYTES).toString("hex");
 }
 
 exports.generateSpanId = generateSpanId;
--- a/lib/propagation.js
+++ b/lib/propagation.js
@@ -1,5 +1,6 @@
 /* global require, exports, Buffer, __dirname */
 const path = require("path"),
+  crypto = require("crypto"),
   pkg = require(path.join(__dirname, "..", "package.json")),
   debug = require("debug")(`${pkg.name}:propagation`),
   schema = require("./schema");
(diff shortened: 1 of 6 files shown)


lib/propagation.test.js
```diff
@@ -16,6 +16,20 @@ let testContext = {
   },
 };
 
+describe("generate span id", () => {
+  test("should create a 8-byte hex id", () => {
+    const regex = /^(([\d||a-f]){16})$/;
```
maplebed, line 21: Is the `||` inside the character class a js regex thing? I would expect to see `[\da-f]` (or `[0-9a-f]`) to represent lowercase hex.

lib/propagation.js
```diff
@@ -11,6 +12,21 @@ exports.REQUEST_ID_HTTP_HEADER = "X-Request-ID";
...
+function generateTraceId() {
+  return crypto.randomBytes(TRACE_ID_BYTES).toString("hex");
+}
+
+function generateSpanId() {
+  return crypto.randomBytes(SPAN_ID_BYTES).toString("hex");
+}
+
+exports.generateSpanId = generateSpanId;
+exports.generateTraceId = generateTraceId;
```
toshok, line 28: even if the plan is to ultimately refactor - following file conventions ++.  these should be moved above their respective functions.

lib/propagation.test.js
toshok: it's not.  it matches and allows `|`.

```
> "012a|5".match(/[\d||a-f]+/)
[ '012a|5', index: 0, input: '012a||5', groups: undefined ]
```

lib/propagation.js
katiebayes: Interesting, I'm accustomed to exporting at the bottom of the doc. Moving

lib/propagation.test.js
katiebayes: whoops, fixing

katiebayes: > I associate generating span and trace IDs with the core tracing internals rather than propagation. I think I would move it, though I don’t know the js layout enough to know if that makes sense. I’d love to hear another take.

@maplebed Yeah we don't currently have a trace document, something that's also likely to change imminently. It definitely didn't feel appropriate to stick it in the libhoney doc, which is where those ids are currently created, so I picked the existing document that felt the closest. I didn't want to expand the scope of this change into any refactoring, but I'll also admit I moved it around a couple times before finally calling it as a "for now." That being said, if you'd like me to put it into a new trace.js document, that's cool with me.

katiebayes:
> I also wouldn't put them in propagation.js, since we aren't planning to tailor trace id _generation_ depending on interop, are we? likewise, I also wouldn't expect them to be used anywhere but the place where startTrace/startSpan live.
> 
> all that said, I don't care too strongly about it.

@toshok this move is actually to bring the formats in line to be friendly with each other so we don't have to tailor them unless there's a specific user input to do so. It's going to continue to be essentially, if there's a trace id, use that, if not, make a new one (in an OTel-consistent type).

lib/propagation.test.js
maplebed: I think to go with the minimal change, I would leave them where they are in libhoney until we’re ready to refactor out a specific trace document. I get not wanting to pull out the trace document right now as it’s a much bigger refactor - that seems like the right choice.

toshok:
I mean, I’d like to see further proposed changes before commenting further, but imo It doesn’t make sense to include fallback trace/span Id generation in the code that extracts/injects. The fallback behavior won’t be tuned by either of those things (since it’s what happens when we don’t have anything injected, and we can’t rely on extraction to provide the fallback impl.)

lib/propagation.js
```diff
@@ -11,6 +12,21 @@ exports.REQUEST_ID_HTTP_HEADER = "X-Request-ID";
 const VERSION = "1";
 exports.VERSION = VERSION;
 
+// id generation
+exports.generateSpanId = generateSpanId;
+exports.generateTraceId = generateTraceId;
```
toshok, line 17: above the _respective_ functions.  look at the way the other functions in the file are done.

katiebayes: @toshok there are several different approaches to this pattern throughout the beeline, and these two functions are uniquely same-but-not. Is this a blocking request?

katiebayes: Nevermind, moved into the libhoney.js file, export no longer needed.

katiebayes:
sorry fixing tests, 10 mins

PR #217, opened 2020-07-08
Refactor header parsers

katiebayes:
Title sums it up! This PR does ~~three~~ two main things: 
- pulls trace header parsers into separate files
- ~~generalizes the marshal and unmarshal functions. Currently just calling the honeycomb parsers (along the lines of keeping functionality the same here)~~ specifies marshal and unmarshal functions separately for aws and honeycomb. The previous functions are still available and wrap the honeycomb functions for backwards compatibility.
- ~~adds an OTel parser for availability in a subsequent PR, definitely willing to bump this change over but I didn't want to change actual functionality in this particular merge~~ pulled this out to a separate PR

Mostly echoing the changes in this [merged PR to the golang beeline](https://github.com/honeycombio/beeline-go/pull/111/files). I even aimed to carry over the function names for consistency and easier documentation.

(don't mind the long list of commits here, I played around a lot on this first one)

Open to any thoughts/suggestions/feedback! One thing I'm majorly uncertain about is where would make the most sense to declare the trace header keys for their respective trace providers. They were previously all in the propagation.js doc that I've broken up into smaller files now. They could all go in index, they're currently declared in the relevant parser file and still exported from propagation/index.js. These currently act as exact string matches for determining which parser to use, so may be a moot point in the near future.


lib/instrumentation/express.js
```diff
@@ -210,7 +210,6 @@ let instrumentExpress = function(express, opts = {}) {
     const app = express();
     // put our middleware in the chain at the start
     app.use(getMagicMiddleware({ userContext, traceIdSource, parentIdSource, packageVersion }));
-
```
toshok, line 213: is vscode removing all these blank lines?  this is annoying.

lib/propagation/w3c_context.js
```diff
@@ -0,0 +1,13 @@
+/* global require, exports */
```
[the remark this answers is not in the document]
katiebayes: Yeah this one isn't used yet and you make a good point that it's not even totally complete, there's more to incorporate about the context structure. I'll bump this to the PR that actually brings in OTel parsing.

package.json
```diff
@@ -37,6 +37,7 @@
     "husky": "^1.3.1",
     "jest": "^24.9.0",
     "jest-in-case": "^1.0.2",
+    "lerna": "^3.22.1",
```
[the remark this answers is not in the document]
katiebayes: erg sorry. Open telemetry uses lerna and this is probably from messing around with that previously.

lib/propagation/aws.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  if (!parentSpanId) {
+    parentSpanId = traceId;
+  }
+
+  return {
+    traceId,
+    parentSpanId,
+    source: `${AMAZON_TRACE_HTTP_HEADER} http header`,
```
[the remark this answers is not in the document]
toshok: mind you, if we move to an extractor model where the extractor function takes a request, it would make sense to add this `source`-setting there.  but it doesn't make sense here.

lib/propagation/honeycomb.js
```diff
@@ -27,8 +26,11 @@ exports.VERSION = VERSION;
 //
 // ex: X-Honeycomb-Trace: 1;trace_id=weofijwoeifj,parent_id=owefjoweifj,context=SGVsbG8gV29ybGQ=
 
-exports.marshalTraceContext = marshalTraceContextv1;
-function marshalTraceContextv1(context) {
+// marshalHoneycombTraceContext takes a honeycomb context object
+// and returns a serialized honeycomb trace header
+exports.marshalHoneycombTraceContext = marshalHoneycombTraceContext;
```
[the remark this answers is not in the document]
katiebayes: It does (I think), but I also like with the suggestion you're making here. This also looks more in line with how we handle event objects and beeline methods, I'll change that over.

lib/instrumentation/express.js
katiebayes: I think that's coming from the precommit hook, which runs some linters. Sorry, I'll try to revert those changes; it's likely running the linter in these files because I committed a couple changes here testing things out

katiebayes: Oh uh...maybe have also just been me. Whoops

lib/propagation/aws.js
```diff
@@ -0,0 +1,36 @@
...
+    if (name === "Root") {
+      traceId = value;
+    } else if (name === "Parent") {
+      parentSpanId = value;
+    }
+  }
+
+  if (!traceId) {
+    // if we didn't even get a 'Root=' clause, bail.
+    return;
```
katiebayes, line 20: switched this to undefined since the fallback to an empty object is still in trace-utils. The tests actually called this out for me, if this returns undefined we can continue to check against truthiness for whether to insert a `source` field.

katiebayes:
Looking much cleaner! Let me know what you think, re-requesting...

lib/instrumentation/trace-util.js
```diff
@@ -21,6 +21,16 @@ const getValueFromHeaders = (requestHeaders, headers) => {
   return undefined;
 };
 
+// adds trace source field to the context and returns the updated context
```
katiebayes, line 24: Pulled this out separately to combine the field-formatting in one spot, but purely preferential and inspired by feedback on the open libhoney PR. Refactor all the logic things!

lib/propagation/honeycomb.js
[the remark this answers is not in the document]
toshok: wow, it does.  I wouldn't have expected the gophers to have been okay with that :)

lib/api/index.js
```diff
@@ -49,12 +51,24 @@ module.exports = {
     return tracker.getTracked();
   },
 
-  TRACE_HTTP_HEADER: propagation.TRACE_HTTP_HEADER,
-  AMAZON_TRACE_HTTP_HEADER: propagation.AMAZON_TRACE_HTTP_HEADER,
   REQUEST_ID_HTTP_HEADER: propagation.REQUEST_ID_HTTP_HEADER,
-  marshalTraceContext: propagation.marshalTraceContext,
+  marshalTraceContext: honeycomb.marshalTraceContext,
```
katiebayes, line 55: Quick update here, I moved this back out of the vendor objects because we don't actually have vendor-specific propagation yet; everything will still propagate in the honeycomb trace format

lib/instrumentation/trace-util.js
```diff
@@ -33,52 +43,29 @@ exports.getTraceContext = (traceIdSource, req) => {
...
-        if (!parsed) {
-          return {};
-        }
-        return Object.assign({}, parsed, {
-          source: `${header} http header`,
-        });
+      //honeycomb trace header
+      case api.honeycomb.TRACE_HTTP_HEADER: {
+        header = api.honeycomb.TRACE_HTTP_HEADER;
+        parsed = api.honeycomb.unmarshalTraceContext(value);
```
toshok, line 52: the `api.honeycomb.` is making me want this code to not actually depend on the exported api at all and instead require the propagations themselves.  but if we end up in an otel-like extractor world this code will be going away anyway, so...

lib/propagation/honeycomb.js
```diff
@@ -27,8 +26,11 @@ exports.VERSION = VERSION;
 //
 // ex: X-Honeycomb-Trace: 1;trace_id=weofijwoeifj,parent_id=owefjoweifj,context=SGVsbG8gV29ybGQ=
 
-exports.marshalTraceContext = marshalTraceContextv1;
-function marshalTraceContextv1(context) {
+// marshalTraceContext takes a trace context object
+// and returns a serialized honeycomb trace header
+exports.marshalTraceContext = marshalTraceContext;
+
+function marshalTraceContext(context) {
```
toshok, line 33: didn't notice before - why drop the `v1`?  It provides a pretty nice correspondence between versioned marshal and unmarshal functions.

lib/propagation/index.js
```diff
@@ -0,0 +1,25 @@
+/* global require, exports */
+const honeycomb = require("./honeycomb");
+exports.aws = require("./aws");
```
toshok, line 3: I meant just to move the require itself - split up the require from the export.

```
const aws = require("./aws);
```

here, and then the honeycomb/aws exports together in either place:

```
exports.honeycomb = honeycomb;
exports.aws = aws;
```

lib/propagation/propagation.test.js
```diff
@@ -1,9 +1,10 @@
 /* global require describe test expect */
-const propagation = require("./propagation"),
-  schema = require("./schema"),
-  Span = require("./api/span"),
+const propagation = require("."),
+  schema = require("../schema"),
+  Span = require("../api/span"),
   cases = require("jest-in-case");
 
+const { honeycomb } = propagation;
```
toshok, line 7: if we're keeping the deprecated `marshalTraceContext` function in propagation/index.js to prevent an api break, why aren't we also keeping `VERSION`?

package-lock.json
```diff
@@ -196,9 +196,9 @@
       },
       "dependencies": {
         "minimist": {
-          "version": "1.2.0",
```
toshok, line 199: as much as I applaud bumping versions of deps, I'd like to keep those commits separate from others.  Separate PR?

lib/instrumentation/trace-util.js
katiebayes: I will say I kind of enjoy the accountability (right word?) of making myself pull these things in from the API. It helps me remember to add them.

lib/propagation/honeycomb.js
katiebayes: v1 is still there! This pattern should be consistent with what we had previously. `unmarshalTraceContext` splits out the version and calls, `unmarshalTraceContextv1` which does the actual parsing

lib/propagation/propagation.test.js
katiebayes: I don't seen any reference to `VERSION` in the api, but maybe we could add it to propagation/index.js to be safe?

toshok: we don't generate a bundle that hides exports.  it's unsafe to remove _anything_ from exports, from anywhere in the package, without a major bump.  +1 to putting it in propagation/index.js

lib/propagation/honeycomb.js
toshok: v1 is _not_ there - `marshalTraceContext` should be `marshalTraceContextv1` to match the unmarshal version.  it was dropped in this PR.

The _export_ is called `marshalTraceContext`.  the original in propagation.js was this:

```
exports.marshalTraceContext = marshalTraceContextv1;

```

PR #220, opened 2020-07-09
Add otel support

katiebayes:
Adding w3c trace hooks. To separate out any breaking change, I'm opting to add without enabling the hooks internally, and will remove the auto-detection (switch statement) logic in a subsequent one with a major version bump. This PR should not alter the behavior of the beeline.

- otel header gets parsed into honeycomb context object
- propagation hook for serializing opentelemetry header and injecting into current context for outbound requests
- tests that w3c hooks work as expected in both directions

### DIFF
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -9,7 +9,7 @@ const path = require("path"),
   LibhoneyImpl = require("./libhoney"),
   MockImpl = require("./mock");
 
-const { honeycomb, aws, otel } = propagation;
+const { honeycomb, aws, w3c } = propagation;
 
 let apiImpl;
 
@@ -69,9 +69,9 @@ module.exports = {
     unmarshalTraceContext: aws.unmarshalTraceContext,
   },
 
-  otel: {
-    TRACE_HTTP_HEADER: otel.TRACE_HTTP_HEADER,
-    unmarshalTraceContext: otel.unmarshalTraceContext,
+  w3c: {
+    TRACE_HTTP_HEADER: w3c.TRACE_HTTP_HEADER,
+    unmarshalTraceContext: w3c.unmarshalTraceContext,
   },
 
   startTrace(metadataContext, withTraceId, withParentSpanId, withDataset) {
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -9,7 +9,7 @@ const path = require("path"),
   LibhoneyImpl = require("./libhoney"),
   MockImpl = require("./mock");
 
-const { honeycomb, aws } = propagation;
+const { honeycomb, aws, otel } = propagation;
 
 let apiImpl;
 
@@ -67,6 +67,11 @@ module.exports = {
   aws: {
     TRACE_HTTP_HEADER: aws.TRACE_HTTP_HEADER,
     unmarshalTraceContext: aws.unmarshalTraceContext,
+  },
+
+  otel: {
+    TRACE_HTTP_HEADER: otel.TRACE_HTTP_HEADER,
+    unmarshalTraceContext: otel.unmarshalTraceContext,
   },
 
   startTrace(metadataContext, withTraceId, withParentSpanId, withDataset) {
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -72,6 +72,7 @@ module.exports = {
   w3c: {
     TRACE_HTTP_HEADER: w3c.TRACE_HTTP_HEADER,
     unmarshalTraceContext: w3c.unmarshalTraceContext,
+    marshalTraceContext: w3c.marshalTraceContext,
   },
 
   startTrace(metadataContext, withTraceId, withParentSpanId, withDataset) {
--- a/lib/api/libhoney.js
+++ b/lib/api/libhoney.js
@@ -105,12 +105,14 @@ module.exports = class LibhoneyEventAPI {
   startTrace(metadataContext, traceId, parentSpanId, dataset) {
     const id = traceId || generateTraceId();
 
+    // initialize an async hook
     tracker.setTracked({
       id,
       traceContext: {},
       stack: [],
       dataset: dataset || this.defaultDataset,
     });
+    // and create a new span to populate the first span in the stack array
     return this.startSpan(metadataContext, undefined, parentSpanId);
   }
 
@@ -120,22 +122,31 @@ module.exports = class LibhoneyEventAPI {
   }
 
   startSpan(metadataContext, spanId = generateSpanId(), parentId = undefined) {
+    // fetch the currently active async hook
+    // "context" here is essentially an "execution context"
     let context = tracker.getTracked();
+
     if (!context) {
       // valid, since we can end up in our instrumentation outside of requests we're tracking
       this.askForIssue("no context in startSpan.");
       return;
     }
+
     if (context.stack.length > 0) {
       if (parentId) {
         debug("parentId supplied when there are already spans.. wrong.");
       }
+
+      // set the parentId to the span id in the most recent entry in the context.stack[] in the async hook
       parentId = context.stack[context.stack.length - 1].payload[schema.TRACE_SPAN_ID];
     }
     if (!parentId) {
       parentId = context.parentId;
     }
 
+    // create new Span object
+    // carry over whatever fields are passed in the first arg
+    // and add the trace id and span id
     const span = new Span(
       Object.assign({}, metadataContext, {
         [schema.TRACE_ID]: context.id,
@@ -143,9 +154,13 @@ module.exports = class LibhoneyEventAPI {
       })
     );
     if (parentId) {
+      // add the parent span id field to the Span object
       span.addContext({ [schema.TRACE_PARENT_ID]: parentId });
     }
+
+    // push the newly created Span into the stack (in the async hook)
     context.stack.push(span);
+
     return span;
   }
 
--- a/lib/api/libhoney.js
+++ b/lib/api/libhoney.js
@@ -122,8 +122,7 @@ module.exports = class LibhoneyEventAPI {
   }
 
   startSpan(metadataContext, spanId = generateSpanId(), parentId = undefined) {
-    // fetch the currently active async hook
-    // "context" here is essentially an "execution context"
+    // fetch the currently active trace
     let context = tracker.getTracked();
 
     if (!context) {
@@ -137,7 +136,7 @@ module.exports = class LibhoneyEventAPI {
         debug("parentId supplied when there are already spans.. wrong.");
       }
 
-      // set the parentId to the span id in the most recent entry in the context.stack[] in the async hook
+      // set the parentId to the span id in the most recent entry in the context.stack[]
       parentId = context.stack[context.stack.length - 1].payload[schema.TRACE_SPAN_ID];
     }
     if (!parentId) {
@@ -158,7 +157,7 @@ module.exports = class LibhoneyEventAPI {
       span.addContext({ [schema.TRACE_PARENT_ID]: parentId });
     }
 
-    // push the newly created Span into the stack (in the async hook)
+    // push the newly created Span into the stack
     context.stack.push(span);
 
     return span;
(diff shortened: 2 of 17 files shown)


lib/instrumentation/trace-util.js
```diff
@@ -47,15 +53,21 @@ exports.getTraceContext = (traceIdSource, req) => {
...
-      case api.aws.TRACE_HTTP_HEADER: {
-        header = api.aws.TRACE_HTTP_HEADER;
-        parsed = api.aws.unmarshalTraceContext(value);
+      case aws.TRACE_HTTP_HEADER: {
+        header = aws.TRACE_HTTP_HEADER;
+        parsed = aws.unmarshalTraceContext(value);
+        return _includeSource(parsed, header);
+      }
+
+      case otel.TRACE_HTTP_HEADER: {
```
katiebayes, line 68: We can skip this if we don't want to add automatic detection for otel. I'd like to keep the removal of this switch statement treatment in favor of config override of the honeycomb default, as out of scope for this PR.

lib/instrumentation/trace-util.js
```diff
@@ -32,30 +32,43 @@ function _includeSource(context, key) {
...
+            otel.TRACE_HTTP_HEADER,
+          ]
         : [traceIdSource];
     let valueAndHeader = getValueFromHeaders(req.headers, headers);
 
     if (!valueAndHeader) {
       return {};
     }
     let { value, header } = valueAndHeader;
+    value = "00-c998e73e5420f609-01";
```
katiebayes, line 52: sorry that's just a test, shouldn't have commited

lib/propagation/propagation.test.js
```diff
@@ -79,6 +79,74 @@ cases(
   ]
 );
 
+cases(
```
katiebayes, line 82: Tests still a WIP, working on finishing it up today!

lib/instrumentation/trace-util.js
```diff
@@ -32,10 +32,11 @@ function _includeSource(context, key) {
 }
 
 exports.getTraceContext = (traceIdSource, req) => {
+  const { honeycomb, aws } = api;
```
katiebayes, line 35: Just destructuring this to clean up where it's used below

lib/propagation/util.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-env node */
```
katiebayes, line 1: I added this file to centralize the structure we're depending on to find the current active trace, but I'm open to other ideas (that don't introduce circular dependencies that break things...not that I did that or anything)

lib/api/index.js
```diff
@@ -69,6 +69,12 @@ module.exports = {
     unmarshalTraceContext: aws.unmarshalTraceContext,
   },
 
+  w3c: {
```
katiebayes, line 72: this could wait for the next PR, could stay here, these are being exported in the api but not used anywhere (yet)

lib/instrumentation.js
```diff
@@ -102,6 +103,9 @@ const instrumentLoad = (exports.instrumentLoad = (mod, loadRequest, parent, opts
 
   let packageVersion = getPackageVersion(loadRequest, parent);
   debug(`loading instrumentation for ${loadRequest}@${packageVersion}`);
+
+  // set instrumentation as the instrumentation function i.e. instrumentHTTP
+  // and require it
```
katiebayes, line 108: adding a comment for my own future reference :)

lib/propagation/honeycomb.js
```diff
@@ -32,7 +32,7 @@ exports.marshalTraceContext = marshalTraceContextv1;
 
 function marshalTraceContextv1(context) {
   let traceId = context.id;
-  let spanId = context.stack[context.stack.length - 1].payload[schema.TRACE_SPAN_ID];
+  let spanId = util.currentSpanId(context);
```
katiebayes, line 35: since this path is in two places now and depends on a structure that could change at some point, it seemed appropriate to pull it out to a util

lib/propagation/w3c.js
```diff
@@ -0,0 +1,92 @@
...
+function marshalTraceContextv1(context) {
+  let spanId = util.currentSpanId(context);
+
+  // find traceFlags setting or fallback to 1 (sampled)
+  let traceFlags = context.traceContext.traceFlags || 1;
+
+  // if traceFlags is truthy and a number, pass it along
+  // this will cover the "00" and future upgrades
+  // undefined will fall back to 1
+  if (
```
katiebayes, line 83: I'm on the fence here. Right now in w3c traces, only one flag is supported (sampling) so traceFlags will either be "01" or "00". The fallback and default should be "01" similar to our default in the beeline to not sample (i.e. sample rate of 1). A totally missing traceFlags, which would be the case for a honeycomb root, should populate the header with "01".

It would be more succinct and foolproof to test for a truthy `traceFlags === 0" and set that, but then we would have to come back and update when and if additional flags get support. As a first pass, I have this checking for a non-1 number and propagating that. Definitely open to other ideas.

lib/instrumentation/trace-util.js
katiebayes: Opting not to add w3 to this switch statement, see PR summary above

lib/propagation/w3c.js
```diff
@@ -0,0 +1,92 @@
...
+
+function marshalTraceContext(context) {
+  return marshalTraceContextv1(context);
+}
+
+function marshalTraceContextv1(context) {
+  let spanId = util.currentSpanId(context);
+
+  // find traceFlags setting or fallback to 1 (sampled)
+  let traceFlags = context.traceContext.traceFlags || 1;
```
toshok, line 78: `.traceContext` here needs to change imo.  the method name is written such that the assumption is that `context` _is_ a trace context.  having a field on it also named `traceContext`...

lib/propagation/propagation.test.js
```diff
@@ -79,6 +79,107 @@ cases(
...
+  "unmarshaling w3c headers",
+  opts => expect(w3c.unmarshalTraceContext(opts.contextStr)).toEqual(opts.value),
+  [
+    {
+      name: "v00, with parent, with span, with traceflags",
+      contextStr: "00-7f042f75651d9782dcff93a45fa99be0-c998e73e5420f609-01",
+      value: {
+        parentSpanId: "c998e73e5420f609",
+        traceId: "7f042f75651d9782dcff93a45fa99be0",
+        customContext: {
```
toshok, line 134: `customContext` is propagated user-instrumentation, not this, imo.

lib/propagation/w3c.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  return unmarshalTraceContextv1(header);
+}
+
+// v1 here refers to honeycomb, the context structure
+function unmarshalTraceContextv1(header) {
+  // pull out required fields, ... for all optional
+  // using spread operator defaults all values to undefined for invalid header values
+  const { traceId, spanId, traceFlags, ...optionalFields } = { ...parseTraceParent(header) };
```
toshok, line 43:
```suggestion
  const { traceId, spanId, traceFlags, ...optionalFields } = parseTraceParent(header);
```

if we're destructuring, we don't also need to clone the return value.

lib/propagation/w3c.js
```diff
@@ -0,0 +1,45 @@
...
+    context: {
+      traceFlags,
+    },
+  };
+
+  // if optionalFields is truthy
+  // and there's something in there
+  if (optionalFields && Object.keys(optionalFields).length !== 0) {
+    // add the optional fields to the context field in the target object (spanContext)
+    addToContext(spanContext.context, optionalFields);
```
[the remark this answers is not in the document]
toshok: hm, yeah see what I said below about putting tracing system-level stuff in customContext.  Think it's a bad fit.

lib/propagation/propagation.test.js
katiebayes: yeah was trying to get this to fit into existing context and threw it in there. Do you think new field? Unique field one level up?

lib/propagation/w3c.js
katiebayes: The spread/clone here gives us the fallback to return `traceId: undefined` for instance instead of the return from parseTraceParent returning undefined (which will break the destructuring). Just thinking how else to work around this. `parseTraceParent` will return `null` if something is found invalid or missing. We could nix the destructure statement altogether, but the spread is really nice for pulling out those extra fields that may or may not be there.

katiebayes: Oh we could test for a null return first, and then move on to this?
```
const parsed = parseTraceParent(header);
if (!parsed) {
  return;
}

const { traceId, spanId, traceFlags, ...optionalFields } = parsed;
```

katiebayes: Yup this passes, I'll push up the change so you can check it out

lib/propagation/w3c.js
katiebayes: I don't disagree, but this seems like potentially a bigger structural change. Currently the `startTrace` function returns an object with `id` (traceId assigned to this field) and `traceContext: {}`. `addTraceContext` in libhoney.js fetches this object from the async hook and adds fields inside `context.traceObject`. Do you want maybe a separate PR to move some of that around before this one?

katiebayes: @toshok a lot of this is kind of still evolving. On second thought, the `context` we're referring to here is really more accurately a _propagation_ context. It's just the currently active async hook, whatever's being tracked. This is the closest thing in golang: https://github.com/honeycombio/beeline-go/pull/118/files#diff-c74a48081d4d8360b4a570379a806b79R510

FWIW this is consistent with what we currently have in honeycomb propagation, where `contextToSend` gets assigned `context.traceContext` and tacked on. The major difference is that an opentelemetry propagation header can't support the open-format serialized, it's either sampled or not currently.

lib/propagation/w3c.js
katiebayes: After discussion, we are actually not going to be propagating the `traceFlags` field right now and will hardcode to "01". As sampling evolves in open telemetry, we will revisit

lib/api/libhoney.js
```diff
@@ -159,13 +158,32 @@ module.exports = class LibhoneyEventAPI {
...
+      dataset: parentContext.dataset,
+      stack: [],
+    };
+
     if (parentContext.stack.length > 0) {
-      parentId = parentContext.stack[parentContext.stack.length - 1].payload[schema.TRACE_SPAN_ID];
+      const latestSpan = parentContext.stack[parentContext.stack.length - 1];
+      newContext.parentId = latestSpan.payload[schema.TRACE_SPAN_ID];
+
+      if (latestSpan.payload.tracestate) {
```
katiebayes, line 174: we're aiming to propagate the `tracestate` header unchanged. I did this pretty quickly but I'm still thinking on this. Ideally it would all be part of the same hook.

lib/propagation/w3c.js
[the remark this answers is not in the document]
katiebayes: We're going to cut this out of this PR completely and just ensure these hooks will propagate the trace id and span id in a usable way. The next PR will cover implementation and will include where to put these additional fields for propagation; some, like traceFlags, we will be dropping for the time being.

lib/api/libhoney.js
katiebayes: I'm going to handle this in a separate PR that allows implementation of these hooks via configuration. Should have a draft up shortly but let's keep this one to marshal/unmarshal and propagating trace id and span id only.

katiebayes:
I'm going to hold off for now on the span API because there are a couple questions to be answered there, mainly should it match the context object we parse headers to, or should we fetch the entire payload, or the whole Span object including timestamps. But I agree that it's a good thing to do in the future to change over to fetching a span.id in the util.

lib/instrumentation.js
```diff
@@ -77,6 +77,7 @@ function getPackageVersion(request, options) {
 }
 
 const instrumentLoad = (exports.instrumentLoad = (mod, loadRequest, parent, opts = {}) => {
+  // loadRequest is a string instrumentation name i.e. express, http
```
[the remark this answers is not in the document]
katiebayes: I saw that it's passed to a require, I think this comment was more that I couldn't tell _what_ was being passed to the require, so figured I'd jot it down in case it was unclear to anyone else. But, point taken and removed :)

PR #223, opened 2020-07-17
Add response context to http/s instrumentation

loganrdavis-ns8:
* Validated locally to see context correctly added to http/s spans.

It would be very beneficial to have some context around these outbound calls in order to determine if the call was successful in logging.

### DIFF
--- a/lib/instrumentation/http.js
+++ b/lib/instrumentation/http.js
@@ -63,6 +63,13 @@ let instrumentHTTP = (http, opts = {}) => {
         },
         span => {
           let wrapped_cb = function(res) {
+            span.addContext({
+              "response.http_version": res.httpVersion,
+              "response.status_code": res.statusCode,
+              "response.content_length": res.headers["content-length"],
+              "response.content_type": res.headers["content-type"],
+              "response.content_encoding": res.headers["content-encoding"],
+            });
             api.finishSpan(span, "request");
             if (callback) {
               return callback.apply(this, [res]);
--- a/lib/instrumentation/http.test.js
+++ b/lib/instrumentation/http.test.js
@@ -19,6 +19,9 @@ beforeAll(done => {
       api.TRACE_HTTP_HEADER,
       req.headers[api.TRACE_HTTP_HEADER.toLowerCase()] || "missing"
     );
+    res.setHeader("Content-Type", "application/json");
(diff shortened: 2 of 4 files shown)


lib/instrumentation/http.js
```diff
@@ -63,6 +63,11 @@ let instrumentHTTP = (http, opts = {}) => {
         },
         span => {
           let wrapped_cb = function(res) {
+            api.addContext({
+              "response.http_version": res.httpVersion,
+              "response.status_code": res.statusCode,
+              "response.headers": res.headers,
```
toshok, line 69: hm, not sure we want all headers added here.  server responses could include things that users don't want sent.  @paulosman do we have some idea of how to address this in other beelines?

loganrdavis-ns8: I figured that'd be a bit controversial. We currently add this context in an interceptor for incoming requests (we filter out sensitive information) but we'd also like this for returned outbound calls.

maplebed: The current go beeline wrapper for outbound requests only includes a[ very limited set ](https://github.com/honeycombio/beeline-go/blob/main/wrappers/hnynethttp/nethttp.go#L166-L178)of returned headers. In general, we've had many very poor experiences trying to slurp up all headers or all query parameters (as an example, [my comment](https://github.com/honeycombio/beeline-go/pull/112#discussion_r445980285) on another recent PR adding support for a go web framework). The middle ground I've thought would be nice is to make it easily configurable in some way. Maybe overload the propagation work that's coming? Or perhaps add another method of an optional callback to give you a place to add fields? Or we might discover that's not necessary because the calling code can add them just as easily. That same gingonic PR had an interesting suggestion on how to allow the caller to provide a list of headers that would be nice to include, but I don't think I have an example of that config.

Absent a desire to add fancy configuration, I suggest grabbing only the small set of headers that matches the go beeline (or a different but equally small set if you have some in mind) and opening a different issue to discuss how to make it more flexible.

loganrdavis-ns8: Great explanation @maplebed and I totally agree. I think for now It's best to simply match the Go Beeline so that a further implementation can be discussed and if desired, applied across repos. I will update to include tests and the Go set of headers.

PR #224, opened 2020-07-17
replace options.method with combinedOptions.method

katiebayes:
Tiny tweak: this was causing requests to reject if the incoming request used the 2-arg format, since "options" may be undefined in that situation.

### DIFF
--- a/lib/instrumentation/http.js
+++ b/lib/instrumentation/http.js
@@ -58,7 +58,7 @@ let instrumentHTTP = (http, opts = {}) => {
         {
           [schema.EVENT_TYPE]: "http",
           [schema.PACKAGE_VERSION]: opts.packageVersion,
-          [schema.TRACE_SPAN_NAME]: options.method || "GET",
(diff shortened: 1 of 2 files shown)


toshok:
reject meaning there was an exception thrown?  this should cause a test failure, no?

katiebayes:
I've made a couple attempts but I'm blanking on what a good test for the actual thing that is breaking would look like. It throws an exception on a child span that then makes another call with the 2-arg format. So needs to be active trace, with a child process that has a `GET` with 2 args. Pretty typical `TypeError: Cannot read property 'method' of undefined`.

I'm going to merge this and keep thinking on it since I'm spinning my wheels a bit, also open to ideas. I'll figure something out before the end of the week.

toshok:
`options` is defined if you give it two args, `options` (it's just either an object or a callback - if it's a callback, `options.method` will be undefined but there won't be a TypeError.)

This test would have exercised it:

```
test("url as a string, 1 argument", done => {
  tracker.setTracked(newMockContext());
  const req = http.get("http://localhost:9009");
  req.on("response", done);
});
```

PR #225, opened 2020-07-17
check for custom context prefix before adding

### LINKED ISSUE #207: custom contexts can get multiple "app" prefixes
Let's imagine I have 3 services using express that depend on each other: A -> B -> C
Within A I call `beeline.addTraceContext({ foo: 'bar' })`. This custom trace context is then propagated to B and then C.

#### Expected
All of my spans have `app.foo: bar`

#### Actual
The span for A contains `app.foo: bar`.
The span for B contains `app.app.foo: bar`.
The span for C contains `app.app.app.foo: bar`.

katiebayes:
Closes #207 

On prefixing custom context fields with `.app` before adding to the span payload, this adds a check for the prefix to avoid adding it multiple times

### DIFF
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -25,7 +25,7 @@ function mapFieldsToApp(map) {
     }
     mapped[schema.customContext(k)] = map[k];
   }
-  console.log(mapped);
+
   return mapped;
 }
 
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -13,11 +13,19 @@ const { honeycomb, aws } = propagation;
 
 let apiImpl;
 
+// incoming custom trace context fields get prefixed
 function mapFieldsToApp(map) {
   let mapped = {};
+
   for (const k of Object.getOwnPropertyNames(map)) {
+    if (!k.startsWith(schema.CUSTOM_CONTEXT_PREFIX)) {
+      mapped[schema.customContext(k)] = map[k];
+    } else {
+      mapped.k = map[k];
+    }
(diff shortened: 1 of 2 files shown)


lib/api/index.js
```diff
@@ -13,11 +13,19 @@ const { honeycomb, aws } = propagation;
 
 let apiImpl;
 
+// incoming custom trace context fields get prefixed
 function mapFieldsToApp(map) {
   let mapped = {};
+
   for (const k of Object.getOwnPropertyNames(map)) {
+    if (!k.startsWith(schema.CUSTOM_CONTEXT_PREFIX)) {
```
katiebayes, line 21: using if/else here since this are mutually exclusive, and opted to check that the key does _not_ start with the prefix already first, since I believe that's going to be more common

lib/schema.js
```diff
@@ -16,4 +16,7 @@ exports.TRACE_SPAN_ID = "trace.span_id";
 exports.TRACE_SERVICE_NAME = "service_name";
 exports.TRACE_SPAN_NAME = "name";
 
-exports.customContext = key => `app.${key}`;
+const CUSTOM_CONTEXT_PREFIX = "app.";
```
katiebayes, line 19: this is a little awkward, but I wanted to be able to pull out this constant to use elsewhere, and also use the same one here in the customContext() function. I landed on declaring it locally and exporting both the string and the function, but definitely open to more elegant ideas here

katiebayes:
This will be moot with my current draft PR, closing!

PR #226, opened 2020-07-18
propagate trace-level fields unchanged, reserving prefixing behavior for api-use only

katiebayes:
Let me know what you think of this, figured I'd take a quick stab while I was thinking of it. For unmarshaling trace headers it really makes more sense to consider anything in the "context" field of a beeline header to be a trace level field and propagate it untouched. In the api, however, let's keep the prefixing so we can identify fields added at beeline-instrumented runtime.

~~I didn't want to add an additional arg and logic to the existing `api.addTraceContext()`~~ after feedback we're totally just going to add another arg to contain any trace level fields parsed from a header, this will enable trace-level fields to pass through untouched, ~~and a new api function seemed unnecessary, so I added a new traceUtil that adds fields to the payload without prefixing. This also allows the api (and functions it wraps) to remain totally untouched.~~

---

`traceContext` -> `spanContext` change is for clarity. A span context is more in line with what we're referring to when we talk about a parsed header. Trace context continues to make sense though for fetching the async hook. 
  - `api.getTraceContext()` fetches the current active trace from the execution (that's staying the same) and now takes an argument for trace-level fields, 
  - ~~`traceUtil.getTraceContext()` --> renamed `traceUtil.getSpanContext()` is calling a parse function on a header. I think it's gotten confusing for these to have the same name~~
  - update on ^ where it's non-breaking, traceContext has been renamed to spanContext when span context is a more accurate term. This is specifically when we are referring to an object derived from a trace header. trace context is more appropriate than custom context in many places, but as this is more of a breaking change, we're mostly sitting on that for now, with comments to help clarify what's going on there

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -242,7 +242,7 @@ beeline.unmarshalTraceContext(traceContext);
 
 Returns an object containing the properties `traceId`, `parentSpanId`, `dataset`, and `customContext` which are the four optional parameters with `startTrace()` above.
 
-`customContext` will include any custom fields propagated from other services, such as with `beeline.addContext()`.
+`customContext` will include any custom fields propagated from other services, such as with `beeline.addTraceContext()`.
 
 example:
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -255,11 +255,11 @@ let trace = startTrace({ name }, traceId, parentSpanId);
 ```
 
 ```javascript
-let { traceId, parentSpanId, customContext } = beeline.unmarshalTraceContext(
+let { traceId, parentSpanId, dataset, customContext } = beeline.unmarshalTraceContext(
   req.header[beeline.TRACE_HTTP_HEADER]
 );
 
-let trace = startTrace({ name }, traceId, parentSpanId, customContext);
+let trace = startTrace({ name }, traceId, parentSpanId, dataset, customContext);
 ```
 
 #### TRACE_HTTP_HEADER
--- a/docs/API.md
+++ b/docs/API.md
@@ -225,7 +225,7 @@ If you're dealing with multiple services (either on the same host or different o
 beeline.marshalTraceContext(beeline.getTraceContext());
 ```
 
-Returns a serialized form of the current trace context (including the trace id and the current span), encoded as a string. The format is documented at https://github.com/honeycombio/beeline-nodejs/blob/master/lib/propagation.js#L16
+Returns a serialized form of the current trace context (including the trace id and the current span), encoded as a string. The format is documented at https://github.com/honeycombio/beeline-nodejs/blob/main/lib/propagation.js#L16
 
 example:
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -27,7 +27,7 @@ In general you're going to be creating spans more often than traces. If you're u
 #### startTrace()
 
 ```javascript
-beeline.startTrace(metadataContext[, withTraceId, withParentSpanId])
+beeline.startTrace(metadataContext[, withTraceId, withParentSpanId, traceContext, withDataset])
 ```
 
 Starts a new local trace and initializes the async context propagation machinery. You _must_ have an active trace for the rest of the API to do anything. If you call `startSpan` when you aren't currently in an trace, an `Error` will be thrown. The instrumentations (which must operate in both trace/non-trace environments) handle this by checking `beeline.traceActive()` and only creating spans if they're within a trace.
@@ -37,7 +37,7 @@ root span is installed as the current span.
 
 Returns a reference to the trace.
 
-For an explanation of `withTraceId` and `withParentSpanId`, see [Interprocess trace propagation](#interprocess-trace-propagation) below.
+For an explanation of `withTraceId`, `withParentSpanId`, and `traceContext` see [Interprocess trace propagation](#interprocess-trace-propagation) below.
 
 example:
 
@@ -240,7 +240,9 @@ console.log(traceContext); // => 1;trace_id=weofijwoeifj,parent_id=owefjoweifj,c
 beeline.unmarshalTraceContext(traceContext);
 ```
 
-Returns an object containing the properties `traceId` and `parentSpanId`, which are the two optional parameters with `startTrace()` above.
+Returns an object containing the properties `traceId`, `parentSpanId`, `dataset`, and `customContext` which are the four optional parameters with `startTrace()` above.
+
+`customContext` will include any custom fields propagated from other services, such as with `beeline.addContext()`.
 
 example:
 
@@ -250,6 +252,14 @@ let { traceId, parentSpanId } = beeline.unmarshalTraceContext(
 );
 
 let trace = startTrace({ name }, traceId, parentSpanId);
+```
+
+```javascript
+let { traceId, parentSpanId, customContext } = beeline.unmarshalTraceContext(
+  req.header[beeline.TRACE_HTTP_HEADER]
+);
+
+let trace = startTrace({ name }, traceId, parentSpanId, customContext);
 ```
 
 #### TRACE_HTTP_HEADER
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -76,8 +76,14 @@ module.exports = {
     marshalTraceContext: w3c.marshalTraceContext,
   },
 
-  startTrace(fields, withTraceId, withParentSpanId, traceContext, withDataset) {
-    return apiImpl.startTrace(fields, withTraceId, withParentSpanId, traceContext, withDataset);
+  startTrace(fields, withTraceId, withParentSpanId, propagatedContext, withDataset) {
+    return apiImpl.startTrace(
+      fields,
+      withTraceId,
+      withParentSpanId,
+      propagatedContext,
+      withDataset
+    );
   },
 
   finishTrace(trace) {
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -69,8 +69,16 @@ module.exports = {
     unmarshalTraceContext: aws.unmarshalTraceContext,
   },
 
-  startTrace(metadataContext, withTraceId, withParentSpanId, withDataset) {
-    return apiImpl.startTrace(metadataContext, withTraceId, withParentSpanId, withDataset);
+  // "fields" can include metadata, root span fields, hardcoded trace names
+  // depending on instrumentation
+  // fields should be an object and any contents will be included in the Span created
+  // a span id passed through will be overridden to create a new Span with new id
+
+  // trace id, parent span id, and dataset can optionally be passed through here
+  // trace id will be generated if absent, parent span id will be stored in the stack
+  // dataset passed directly here will override the beeline configuration
+  startTrace(fields, withTraceId, withParentSpanId, withDataset) {
+    return apiImpl.startTrace(fields, withTraceId, withParentSpanId, withDataset);
   },
   finishTrace(trace) {
     return apiImpl.finishTrace(trace);
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -83,8 +83,8 @@ module.exports = {
   // trace id, parent span id, and dataset can optionally be passed through here
   // trace id will be generated if absent, parent span id will be stored in the stack
   // dataset passed directly here will override the beeline configuration
-  startTrace(fields, withTraceId, withParentSpanId, withDataset) {
-    return apiImpl.startTrace(fields, withTraceId, withParentSpanId, withDataset);
+  startTrace(fields, withTraceId, withParentSpanId, withDataset, traceContext) {
+    return apiImpl.startTrace(fields, withTraceId, withParentSpanId, withDataset, traceContext);
   },
 
   finishTrace(trace) {
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -75,16 +75,8 @@ module.exports = {
     marshalTraceContext: w3c.marshalTraceContext,
   },
 
-  // "fields" can include metadata, root span fields, hardcoded trace names
-  // depending on instrumentation
-  // fields should be an object and any contents will be included in the Span created
-  // a span id passed through will be overridden to create a new Span with new id
-
-  // trace id, parent span id, and dataset can optionally be passed through here
-  // trace id will be generated if absent, parent span id will be stored in the stack
-  // dataset passed directly here will override the beeline configuration
-  startTrace(fields, withTraceId, withParentSpanId, withDataset, traceContext) {
-    return apiImpl.startTrace(fields, withTraceId, withParentSpanId, withDataset, traceContext);
+  startTrace(fields, withTraceId, withParentSpanId, traceContext, withDataset) {
+    return apiImpl.startTrace(fields, withTraceId, withParentSpanId, traceContext, withDataset);
   },
 
   finishTrace(trace) {
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -9,7 +9,7 @@ const path = require("path"),
   LibhoneyImpl = require("./libhoney"),
   MockImpl = require("./mock");
 
-const { honeycomb, aws } = propagation;
+const { honeycomb, aws, w3c } = propagation;
 
 let apiImpl;
 
@@ -69,6 +69,12 @@ module.exports = {
     unmarshalTraceContext: aws.unmarshalTraceContext,
   },
 
+  w3c: {
+    TRACE_HTTP_HEADER: w3c.TRACE_HTTP_HEADER,
+    unmarshalTraceContext: w3c.unmarshalTraceContext,
+    marshalTraceContext: w3c.marshalTraceContext,
+  },
+
   // "fields" can include metadata, root span fields, hardcoded trace names
   // depending on instrumentation
   // fields should be an object and any contents will be included in the Span created
@@ -80,6 +86,7 @@ module.exports = {
   startTrace(fields, withTraceId, withParentSpanId, withDataset) {
     return apiImpl.startTrace(fields, withTraceId, withParentSpanId, withDataset);
   },
+
   finishTrace(trace) {
     return apiImpl.finishTrace(trace);
   },
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -76,13 +76,13 @@ module.exports = {
     marshalTraceContext: w3c.marshalTraceContext,
   },
 
-  startTrace(fields, withTraceId, withParentSpanId, propagatedContext, withDataset) {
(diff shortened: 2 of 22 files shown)


lib/instrumentation/fastify.js
```diff
@@ -97,13 +100,15 @@ const instrumentFastify = function(fastify, opts = {}) {
...
+        spanContext.dataset
       );
 
-      if (traceContext.customContext) {
-        api.addTraceContext(traceContext.customContext);
+      if (spanContext.traceContext) {
+        // if the parsed header contains trace level fields
+        // propagate them without prefixing or changing them
+        traceUtil.addUnchangedContextField(spanContext.customContext);
       }
```
toshok, line 112: to be honest I've never understood why this "provide initial propagated trace context" wasn't one of the args passed to `api.startTrace`.  every other piece we pull out of headers is, why not this?  I have no idea why I didn't roll it in previously.

how about we lose `traceUtil.addUnchangedContextField` (it becomes more api surface we can't change/remove - as it's necessary to write new web framework instrumentation), and instead pass .customContext as another arg to `api.startTrace`

Then `startTrace` can change to something like:

```diff
-startTrace(metadataContext, traceId, parentSpanId, dataset) {
+startTrace(rootSpanContext, traceId, parentSpanId, dataset, propagatedContext = {}) {
    const id = traceId || uuidv4();

    tracker.setTracked({
      id,
-      traceContext: {},
+      traceContext: propagatedContext,
      stack: [],
      dataset: dataset || this.defaultDataset,
    });
```

an argument could be made for keeping the ability to parse/install propagated context at trace creation time as part of the api, but given that we're going to let folks install their own ingest time parse hooks, why not force them to write one of those to return the proper values?

The go beeline's `NewTrace` does things quite a bit differently in this regard (it's passed the headers - something the node beeline can't rely on - different frameworks have different header apis), but it also doesn't require another exported API call to build up the propagated trace context - it's done as part of `NewTrace`.

lib/instrumentation/trace-util.js
```diff
@@ -31,7 +31,7 @@ function _includeSource(context, key) {
   });
 }
 
-exports.getTraceContext = (traceIdSource, req) => {
+exports.getSpanContext = (traceIdSource, req) => {
```
toshok, line 34: breaking change.

lib/instrumentation/trace-util.js
```diff
@@ -76,6 +76,20 @@ exports.getTraceContext = (traceIdSource, req) => {
   }
 };
 
+exports.addUnchangedContextField = map => {
```
toshok, line 79: this is either something we need to expose and support, or this is something that only our own instrumentation has access to.  I'd prefer _not_ to have to export it (as it's a very special case of `addTraceContext`), and having a necessary bit of api that we don't export is also not good.

lib/instrumentation/fastify.js
toshok: given that _every_ arg after the first one is meant to be installed after parsing headers, we can likely change startTrace to be more similar to the go beeline, and:

```
startTrace(rootSpanContext, propagation) 
// where propagation: { traceId, parentSpanId, dataset, propagatedContext, ... }
```

but again with the breaking changes (unless we support both calling sequences.)

lib/propagation/honeycomb.js
```diff
@@ -110,7 +110,7 @@ function unmarshalTraceContextv1(payload) {
   return {
     traceId,
     parentSpanId,
-    customContext,
+    traceContext,
```
toshok, line 113: breaking change here too

katiebayes:
jumping back into this now, sorry for the delay

lib/instrumentation/trace-util.js
katiebayes: ok, reverted this and added some comments to make it clear what's going on there

lib/instrumentation/fastify.js
katiebayes: I definitely prefer adding any existing propagation through unchanged rather than the cycle-through and prefix or don't-prefix. How about we roll with that for now, and aim to bring it more in line with the golang process in an upcoming version bump? We're going to have to push a major version with the configuration changes, not sure if it's better to loop that in, or wait for a future update. Any preference?

katiebayes: Minor nit, I'm not sure `rootSpanContext` is accurate or perhaps is just a little too specific. The current usage of that argument *does* pass metadata through in the express and fastify instrumentations, declaring it as 'fields'. Those fields are not part of a trace or span context per se but do get grouped in with the span when it's added to `Span.payload.stack[]`. I honestly think a simple `fields` argument there and a comment is the way to go here. Span context is a loaded term with a specific definition of what fields would be in there, that I think we need to be very careful with moving forward

lib/api/index.test.js
```diff
@@ -561,10 +561,10 @@ test("distributed tracing linkage is correct", () => {
       [schema.TRACE_SPAN_NAME]: "sub",
     },
     ctx.traceId,
-    ctx.parentSpanId
+    ctx.parentSpanId,
```
katiebayes, line 564: this is closer to what we're actually doing with a parsed header with this change

lib/api/libhoney.js
```diff
@@ -102,18 +102,20 @@ module.exports = class LibhoneyEventAPI {
...
+  // traceContext will include any trace-level fields
+  // can be the context field in a honeycomb header or tracestate in w3c
+  startTrace(fields, traceId, parentSpanId, dataset, traceContext = {}) {
     const id = traceId || generateTraceId();
 
     // initialize an async hook
     tracker.setTracked({
       id,
-      traceContext: {},
+      traceContext,
```
katiebayes, line 114: empty object has been moved to the argument default

lib/instrumentation/fastify.js
toshok: yes.  there are two different uses of context throughout the beeline.  there's "I'm adding context to my spans" - that's rootSpanContext, metadataContext, propgatedContext, and the traceContext that hangs off the _tracked_ trace context.

those would all be better served by getting away from the "context" naming and switching to "fields".

toshok: > I definitely prefer adding any existing propagation through unchanged rather than the cycle-through and prefix or don't-prefix. How about we roll with that for now, and aim to bring it more in line with the golang process in an upcoming version bump? 

I don't understand what you're asking here.

lib/api/index.test.js
toshok: sure, but withDataset can be undefined.  if we're going to specify it here, might as well exercise it in the assertions below (set it in the context, have it marshaled, verify that it affects the sent events in the subservice).

lib/api/libhoney.js
```diff
@@ -102,18 +102,20 @@ module.exports = class LibhoneyEventAPI {
     });
   }
 
-  startTrace(metadataContext, traceId, parentSpanId, dataset) {
+  // fields will be included in a new Span
+  // traceContext will include any trace-level fields
+  // can be the context field in a honeycomb header or tracestate in w3c
```
toshok, line 107: more notes?

lib/instrumentation/express.js
```diff
@@ -83,16 +83,28 @@ const getMagicMiddleware = ({ userContext, traceIdSource, parentIdSource, packag
   res,
   next
 ) => {
-  let traceContext = traceUtil.getTraceContext(traceIdSource, req);
+  // parse incoming trace headers into a useable *span* context
+  let spanContext = traceUtil.getTraceContext(traceIdSource, req);
+
+  // parentTraceId refers to the span id of the parent span
```
toshok, line 89: imo instead of documenting this discrepancy, fix it by making the variable `parentSpanId`.

lib/instrumentation/express.js
```diff
@@ -115,15 +127,13 @@ const getMagicMiddleware = ({ userContext, traceIdSource, parentIdSource, packag
 
   let rootSpan = api.startTrace(
     fields,
-    traceContext.traceId,
-    traceContext.parentSpanId,
-    traceContext.dataset
+    spanContext.traceId,
+    spanContext.parentSpanId,
+    spanContext.dataset,
+    // trace level fields, incl propagated custom context
```
toshok, line 133: I think we don't need another comment saying what this variable is - it's already documented above.

lib/instrumentation/express.js
```diff
@@ -83,16 +83,28 @@ const getMagicMiddleware = ({ userContext, traceIdSource, parentIdSource, packag
   res,
   next
 ) => {
-  let traceContext = traceUtil.getTraceContext(traceIdSource, req);
+  // parse incoming trace headers into a useable *span* context
```
toshok, line 86: lose the emphasis
```suggestion
  // parse incoming trace headers into a span context
```

lib/instrumentation/fastify.js
```diff
@@ -70,16 +70,29 @@ const instrumentFastify = function(fastify, opts = {}) {
 
     app.addHook("onRequest", (request, reply, next) => {
       // start our trace handling
-      let traceContext = traceUtil.getTraceContext(traceIdSource, request);
+
+      // parse incoming trace headers into a useable *span* context
```
toshok, line 74: same comments from express.js in this method

lib/instrumentation/trace-util.js
```diff
@@ -31,6 +31,9 @@ function _includeSource(context, key) {
   });
 }
 
+// this function parses a trace header and returns a SPAN context object
+// may contain some trace-level data, but consider this a span context
+// naming staying traceContext to avoid breaking changes
```
toshok, line 36: if it's holding onto a name to avoid breaking changes, mark it deprecated and suggest the new name (or change) that should be expected in the next major version.  if we don't have a path forward yet, say that.  If it's not enough to deprecate, lose that bit entirely.

lose the emphasis as well, and also "may contain some trace-level data".

basically:

```
// parse a trace header and return object used to populate args to startTrace
// deprecated: in the next major version etc etc etc.
```

(or `s/object/span context object/` if you truly must)

lib/instrumentation/trace-util.test.js
```diff
@@ -75,7 +75,7 @@ describe("getTraceContext", () => {
         expectedContext: {
           traceId: "abcdef",
           parentSpanId: "12345",
-          customContext: undefined,
+          traceContext: undefined,
```
toshok, line 78: this still needs reverting.

lib/instrumentation/express.js
```diff
@@ -83,16 +83,28 @@ const getMagicMiddleware = ({ userContext, traceIdSource, parentIdSource, packag
...
+  // parentTraceId refers to the span id of the parent span
   let parentTraceId = traceUtil.getParentSourceId(parentIdSource, req);
   if (parentTraceId) {
-    traceContext.parentSpanId = parentTraceId;
+    spanContext.parentSpanId = parentTraceId;
+  }
+
+  // if the parsed header contains trace level fields
+  // propagate them without prefixing or changing them
+  // includes custom context from a parent span
```
toshok, line 97:
```suggestion
  // propagate trace level fields as-is, if there are any
```

katiebayes:
> kind of annoyed that a bug fix for not prefixing custom trace fields with `app.` got rolled into this giant bikeshed.

if it helps, this is also going to be very helpful for cleaner propagation of trace-level context, so it was this one or the next one honestly.

lib/instrumentation/express.js
katiebayes: Totally agree, making the change

katiebayes:
Ok, I think I covered everything noted. Let me know what you think of this. I opted to swap the ordering of the dataset and traceContext since dataset is never default-populated.

katiebayes:
Updated from `main`

lib/instrumentation/fastify.js
katiebayes: I was just referring to the breaking change you mentioned, passing a single argument with multiple fields. For now I've added the additional `propagatedContext` as an optional argument to `startTrace`

katiebayes: Just swapped `traceContext` to `propagatedContext` as the argument names, I think that's more clear

lib/api/index.test.js
katiebayes: Agree, but for this situation I opted to just add customContext. I can push up more withDataset tests/scenarios in another PR.

lib/api/index.js
```diff
@@ -75,9 +75,18 @@ module.exports = {
...
-  startTrace(metadataContext, withTraceId, withParentSpanId, withDataset) {
-    return apiImpl.startTrace(metadataContext, withTraceId, withParentSpanId, withDataset);
+  // "fields" can include metadata, root span fields, hardcoded trace names
+  // depending on instrumentation
+  // fields should be an object and any contents will be included in the Span created
+  // a span id passed through will be overridden to create a new Span with new id
+
+  // trace id, parent span id, and dataset can optionally be passed through here
+  // trace id will be generated if absent, parent span id will be stored in the stack
+  // dataset passed directly here will override the beeline configuration
```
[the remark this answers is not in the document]
katiebayes: took this out completely, updated api docs instead

lib/api/libhoney.js
katiebayes: cut this out, just kept one note that both `fields` and `propagatedContext` are objects

docs/API.md
```diff
@@ -240,7 +240,9 @@ console.log(traceContext); // => 1;trace_id=weofijwoeifj,parent_id=owefjoweifj,c
 beeline.unmarshalTraceContext(traceContext);
 ```
 
-Returns an object containing the properties `traceId` and `parentSpanId`, which are the two optional parameters with `startTrace()` above.
+Returns an object containing the properties `traceId`, `parentSpanId`, `dataset`, and `customContext` which are the four optional parameters with `startTrace()` above.
+
+`customContext` will include any custom fields propagated from other services, such as with `beeline.addContext()`.
```
toshok, line 245:
```suggestion
`customContext` will include any custom fields propagated from other services, such as with `beeline.addTraceContext()`.
```

docs/API.md
```diff
@@ -252,6 +254,14 @@ let { traceId, parentSpanId } = beeline.unmarshalTraceContext(
...
 ```
 
+```javascript
+let { traceId, parentSpanId, customContext } = beeline.unmarshalTraceContext(
+  req.header[beeline.TRACE_HTTP_HEADER]
+);
+
+let trace = startTrace({ name }, traceId, parentSpanId, customContext);
+```
+
```
toshok, line 264: I think add `customContext` (and the dataset - positional args means we can't skip that) in the snippet directly above this one.

lib/api/index.js
```diff
@@ -76,9 +76,16 @@ module.exports = {
     marshalTraceContext: w3c.marshalTraceContext,
   },
 
-  startTrace(metadataContext, withTraceId, withParentSpanId, withDataset) {
-    return apiImpl.startTrace(metadataContext, withTraceId, withParentSpanId, withDataset);
+  startTrace(fields, withTraceId, withParentSpanId, propagatedContext, withDataset) {
```
toshok, line 79: bleah, didn't notice this before - this is a breaking change too.  can't add an argument into the middle of the list here.  `propagatedContext` has to go at the end.

lib/api/libhoney.js
```diff
@@ -102,18 +102,18 @@ module.exports = class LibhoneyEventAPI {
     });
   }
 
-  startTrace(metadataContext, traceId, parentSpanId, dataset) {
+  // fields and propagatedContext are both objects
+  startTrace(fields, traceId, parentSpanId, propagatedContext = {}, dataset) {
```
toshok, line 106: same here.  propagatedContext has to go at the end.

lib/api/index.test.js
toshok: we'll need an extra parameter regardless.  might as well add it.

lib/api/index.js
katiebayes: hmm that's tricky then, because `withDataset` can be undefined and `propagatedContext` has a fallback. So with that and adding a new arg, we may have a problem with propagatedContext getting interpreted as withDataset. Check against the type maybe? String is dataset?

toshok: I don't think it's ambiguous or problematic - there's no default for withDataset, so if you want to pass propagatedContext, you must pass `undefined` if there isn't one.

toshok: I'd much prefer if none of our api had optional args in the middle.  there's just no real need for it.

katiebayes: Well technically both of these args are optional :) But sure, we can default dataset to undefined and pass it regardless, that should work

docs/API.md
katiebayes: I know you suggested it already but I think in the next major version we 100% should switch these arguments to a single object so we can nix the positional thing

katiebayes: Ok dataset added in here

lib/api/index.js
katiebayes: Order swapped back

lib/api/libhoney.js
katiebayes: here too

lib/api/index.test.js
katiebayes: Ok added another test and a couple lines to the existing test to check dataset and trace context on honeycomb events.

lib/instrumentation/express.js
```diff
@@ -115,15 +125,12 @@ const getMagicMiddleware = ({ userContext, traceIdSource, parentIdSource, packag
 
   let rootSpan = api.startTrace(
     fields,
-    traceContext.traceId,
-    traceContext.parentSpanId,
-    traceContext.dataset
+    spanContext.traceId,
+    spanContext.parentSpanId,
+    propagatedContext,
+    spanContext.dataset
```
toshok, line 131: these are the wrong order still, I think?  some missing test coverage, methinks.

lib/instrumentation/fastify.js
```diff
@@ -97,15 +110,12 @@ const instrumentFastify = function(fastify, opts = {}) {
       }
       let rootSpan = api.startTrace(
         fields,
-        traceContext.traceId,
-        traceContext.parentSpanId,
-        traceContext.dataset
+        spanContext.traceId,
+        spanContext.parentSpanId,
+        traceContext,
+        spanContext.dataset
```
toshok, line 116: same here.

lib/instrumentation/express.js
katiebayes: huh, sorry I missed these but tests were passing. Doing it right now let's get this one merged and I'll take a look at the tests later

PR #227, opened 2020-07-21
add tests to confirm one-argument get requests will work

katiebayes:
Confirmed this test breaks for the exception I was running into, thanks @toshok!


katiebayes:
Looking into the protocol mismatch now

katiebayes:
I take it back, whoops. Closing this for now and will revisit.

katiebayes:
In a nutshell, it fails the way we would expect, but does not pass the way we would expect. The example I got the exception on was using the fetch API, maybe there's something up there.

PR #228, opened 2020-07-22
add aws marshal function and test

katiebayes:
Realized we need an aws _marshal_ function as well! I mostly was aiming to translate this recent addition to the [go beeline](https://github.com/honeycombio/beeline-go/blob/main/propagation/amazon.go#L17), just translated to javascript!

### DIFF
--- a/lib/propagation/aws.js
+++ b/lib/propagation/aws.js
@@ -37,7 +37,7 @@ function unmarshalTraceContext(header) {
     } else if (name === "Self") {
       parentSpanId = value;
     } else {
-      customContext[name] = value
+      customContext[name] = value;
     }
   }
 
--- a/lib/propagation/aws.js
+++ b/lib/propagation/aws.js
@@ -20,21 +20,22 @@ function marshalTraceContextv1(context) {
     contextToSend = ";" + elements.join(";");
   }
 
-  return `Root=1-${traceId};Self=${spanId}${contextToSend}`;
+  return `Root=1-${traceId};Parent=${spanId}${contextToSend}`;
 }
 
 // unmarshalTraceContext takes a string trace header and returns a context
 exports.unmarshalTraceContext = unmarshalTraceContext;
 
 function unmarshalTraceContext(header) {
-  let traceId, parentSpanId, customContext;
+  let traceId, parentSpanId;
+  let customContext = {};
 
   const split = header.split(";");
   for (const s of split) {
     const [name, value] = s.split("=");
     if (name === "Root") {
       traceId = value;
-    } else if (name === "Self") {
+    } else if (name === "Parent") {
       parentSpanId = value;
     } else {
       customContext[name] = value;
--- a/lib/propagation/aws.js
+++ b/lib/propagation/aws.js
@@ -37,9 +37,7 @@ function unmarshalTraceContext(header) {
     } else if (name === "Self") {
       parentSpanId = value;
     } else {
-      customContext = Object.assign({}, customContext, {
-        [name]: value,
-      });
+      customContext[name] = value
     }
   }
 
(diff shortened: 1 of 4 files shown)


lib/propagation/aws.js
```diff
@@ -1,7 +1,28 @@
...
+// and returns a serialized honeycomb trace header
+exports.marshalTraceContext = marshalTraceContextv1;
+
+function marshalTraceContextv1(context) {
+  let traceId = context.id;
+  let spanId = util.currentSpanId(context);
+  let contextToSend = "";
+
+  if (context.traceContext) {
+    const elements = Object.keys(context.traceContext).map(key => {
```
toshok, line 17: probably worth accepting an undefined `context.traceContext` / supplying a `{}` default, ala `propagation/honeycomb.js`.

katiebayes: good catch, I'll just make that exactly the same as honeycomb

lib/propagation/aws.js
```diff
@@ -30,5 +55,6 @@ function unmarshalTraceContext(header) {
   return {
     traceId,
     parentSpanId,
+    customContext,
```
katiebayes, line 58: this would ideally be `traceContext` in all unmarshal-ers, but since customContext is the one we're using currently I think it's fine to leave this as-is. customContext will get pushed into `Span.traceContext`. If we don't want to do this, we have two options that I can see: one is to add a possible `traceContext` field that can return from an unmarshal in addition to customContext for non-honeycomb providers, the other would be to change the name of this field everywhere.

katiebayes:
Re-requesting because the changes to aws unmarshal are non-negligible. Lemme know what you think of this, see comments for a couple things I'm not sure on.

lib/propagation/aws.js
```diff
@@ -1,20 +1,45 @@
...
-  let traceId, parentSpanId;
+  let traceId, parentSpanId, customContext;
 
   const split = header.split(";");
   for (const s of split) {
     const [name, value] = s.split("=");
     if (name === "Root") {
       traceId = value;
-    } else if (name === "Parent") {
+    } else if (name === "Self") {
```
katiebayes, line 37: Looking at the docs [here](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html) and the go beeline, it looks like we should be looking for `Self` as a load balancer identifying itself, which we would then consider the parent span. This does mean `Parent` will go into trace-level context which is iffy, but also doesn't seem common for ALB specifically.

lib/propagation/aws.js
toshok: if/when we rename things, this should be `traceLevelFields`. We need to lose as many "*Contexts" as possible.  "traceContext" means different things depending on where in the code you are.  and in some places (like `startTrace`) we have to deal with both at the same time - recipe for confusion.

+100 leave as is for now - having unmarshal methods return differently shaped objects is worse than the pain we're dealing with now, imo.

lib/propagation/aws.js
toshok: we don't want `Self` here.  we want `Parent`

https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/patchers/http_p.js#L122

toshok: I mean, maybe we need to handle Self if Parent doesn't exist, but I don't recall seeing that /shrug

lib/propagation/aws.js
```diff
@@ -27,15 +27,19 @@ function marshalTraceContextv1(context) {
...
       parentSpanId = value;
+    } else {
+      customContext = Object.assign({}, customContext, {
+        [name]: value,
+      });
```
toshok, line 42: maybe `customContext = {}` above, then:

```suggestion
    } else {
      customContext[name] = value
```

here?

katiebayes:
I'm actually switching back to the object, because I think returning only `parentSpanId` and `traceId` rather than adding `customContext: {}` when it's undefined is more consistent with how we handle other potentially undefined parsed fields

PR #231, opened 2020-07-24
fix Parent/Self parsing in aws parse and propagation functions, other aws cleanup/enhancement

katiebayes:
Updates to AWS parsing:
- If `Self` is present and `Parent` is not, use `Self` as `parentSpanId`
- If `Parent` is present and `Self` is not, use `Parent` as `parentSpanId`
- If _both_ `Self` and `Parent` are present, set `Self` as `parentSpanId` and ignore `Parent`

Other improvements at @MikeGoldsmith's suggestion (thank you!):
- case insensitivity for header detection
- check for empty string as well as undefined

Related to https://github.com/honeycombio/beeline-go/pull/122 in the go beeline


paulosman:
> aws's own SDKs don't use `Self` for this purpose (they use `Parent`). `Self` is entirely for load balancers (it's the id of the nearest LB span - LB's are the ones that are supposed to update it.) 

Just chiming in to say, compatibility with X-Ray is not our goal. We want to support one use case only for now, and that's being able to connect to a root span that originates from an amazon load balancer. 

I haven't had a chance to look at the PR yet, just wanted to make sure that context was provided.

toshok:
> Just chiming in to say, compatibility with X-Ray is not our goal.

we want to interop with other tracers, right?  none of them use `Self`

paulosman:
> > Just chiming in to say, compatibility with X-Ray is not our goal.
> 
> we want to interop with other tracers, right? none of them use `Self`

We don't, that's not the goal. It's literally just load balancers that we're interested in 😄

toshok:
> We don't, that's not the goal. It's literally just load balancers that we're interested in 😄

Right, this is the problem.  We are choosing an incompatible interpretation of the same header, and calling it the same thing.  This is going to cause confusion.  It also paints us into a corner where if we ever want to fix it to have the correct semantics, it's a breaking change.

A few options here:

1. call it something other than `aws`
2. leave it in `aws` but add it as an opt-in behavior (`useSelfAsParent: true`), with the default being "use parent."
3. add it as a fallback (use `Parent` if present, use `Self` otherwise) - ugh, magic, and occasionally breaking magic.
4. Leave `aws` alone, don't include this incompatible behavior in the beelines at all, and in line with the "be explicit instead of using magic" philosophy for trace interop: document how to write a custom hook that does this.

MikeGoldsmith:
Okay, my thoughts on testing @paulosman did with the [ALB](https://gist.github.com/paulosman/f65a7529b410d9c439c6065a4c0a8fbb) are:
1) `root` is required, will be generated if one is not provided
2) `self` is optional, and is only appended if `root` is already present, it will always generate a new value (ie will not propagate an existing value if set)
2) `Parent` is logically ignored and propagated if present


I agree with @paulosman's assessment that the goal of this propagator is to only work with traces that originate from an Amazon ALB, not traces that are multi-process propagated. There are better trace context formats that you'd be expected to use in place of piggybacking onto the AWS header. We're working to improve the beelines to allow 1+ propagator implementations to run which means we could pick which propagation format is preferred instead of picking just one and making it more complex.

However, I also take @toshok's point that the expectation would be to use `Parent` if it is present in the header, as this would break other tracer implementations that use `Parent` and not `Self`.

As the intention is to only support new trace graphs that originate from an ALB; I propose we use `Parent` if present over `Self`. This would allow use to visualise trace's that originate at ALB and also not break a trace graph if a Parent is present when it's set from by another process.

Quick example (please forgive my JS 😄)
```javascript
let traceId, parentSpanId, customContext;
const split = header.split(";");
for (const s of split) {
  const [name, value] = s.split("=");
  if (name === "Root") {
    traceId = value;
  } else if (name === "Self") {
    // only use self as parent ID if one is not already set
    if (!parentSpanId) {
      parentSpanId = value;
    }
  } else if (name === "Parent") {
    parentSpanId = value;
  } else {
    customContext = Object.assign({}, customContext, {	      customContext = Object.assign({}, customContext, {
      [name]: value,
    });
  }
}

```

---
@katiebayes I think it's also worth noting we do not validate any key pair value to not be empty and maybe we should consider a case-insensitive keys too? I've seen some truly weird stuff happy with headers before. Do you think it's worth adding?

paulosman:
@toshok @MikeGoldsmith @katiebayes whatever we decide should:

- Support load balancer originating traces out-of-the-box 
- Work consistently across all the beelines.

I'd like to make a decision soon so we can move forward with releases. I think I'm okay with "Parse parent unless self is present" but it would require changes to the other beelines to make them consistent, so I want to be sure it provides value. Knowing that supporting interop with X-ray isn't a goal, I'm still unsure who we're trying to support (or is it future-proofing?).

lib/propagation/aws.js
```diff
@@ -20,41 +20,60 @@ function marshalTraceContextv1(context) {
...
-    } else if (name === "Parent") {
-      parentSpanId = value;
-    } else {
-      customContext = Object.assign({}, customContext, {
-        [name]: value,
-      });
-    }
+function unmarshalTraceContext(header = "") {
+  let parentSpanId, customContext;
+  const supportedFields = ["root", "self", "parent"];
```
katiebayes, line 31: identifying the fields that should transform to lowercase for case-insensitive comparison. I do feel we should _not_ transform custom fields, hence this change.

maplebed:
I'd like to pull in some explicit stories to help guide the goals for this change. 

Scenario 1: external traffic comes into an ALB and then lands at a beeline-instrumented service. The goal is to allow the ALB to be the root of the trace. Need: parse `root`, use it as both trace ID and parent ID for the new span in the beeline-instrumented service.

Scenario 2: traffic starting in a beeline-instrumented service calls out to another beeline-instrumented service via an ALB (aka `svcA` -> `ALB` -> `svcB`). The goal is to have the ALB visible in the trace. Need: `svcA` must send `root` and `self`; the ALB will log both (to be interpreted by honeyelb), keep `root` and overwrite `self`. The trace ID, span IDs, and parent IDs of each segment are available to use. The receiving service (`svcB`) will need to parse both `root` and `self` and set the trace ID to `root` and `self` becomes the `svcB`'s parent ID.

I don't have a scenario that involves `parent` in the AWS header so I don't have any framework in which to understand what we would need to do with it.  (see below for a guess.)

It's also worth noting that what we're providing here is the "works out of the box for the simple case" behavior while also enabling the people using the beeline to write any more complicated logic themselves. If someone does manage to come up with a way that `parent` influences the propagation of IDs and has a clear understanding of how to make wire that through to our trace waterfall, there is nothing stopping them from implementing that, and we can use it to guide future changes. 


Ok, @toshok you did link to the AWS Java SDK for working with Open Telemetry, and said "it sends parent". Assuming we could get this to play in our trace, what would it look like? Let's assume scenario 2 above, but when `svcA` (the otel java sdk one) is configured using the Honeycomb exporter and sends us a trace ID and span ID for the span calling into a load balancer. It also sets the outgoing header with `root` and `parent`. The ALB logs both `root` and `parent` for honeytail's later reading, and sends a header to our beeline-instrumented service that has `root`, `self`, and `parent`.  The beeline parses `root` as the trace ID, takes `self` as the parent ID (representing the ALB), and generates a new span ID. The trace is complete, according to the _beeline_. In this scenario, `honeytail` would need to be the one that understands that Parent represents the span calling into the ALB, but the beeline's parser still only needs root and self. 

I'd love to see this land matching the [go beeline's logic](https://github.com/honeycombio/beeline-go/blob/main/propagation/amazon.go#L68-L81) - look for root and self, treat everything else as optional extra parameters.

lib/propagation/aws.js
```diff
@@ -20,41 +20,60 @@ function marshalTraceContextv1(context) {
...
+      // return lowercase keys for parsable fields to allow case insensitivity
+      if (supportedFields.includes(key.toLowerCase())) {
+        return { [key.toLowerCase()]: value };
+      }
+      return { [key]: value };
+    })
+    .reduce((acc, obj) => Object.assign(acc, obj));
+
+  // customContext remain undefined unless there are fields
+  if (Object.keys(rest).length !== 0) {
```
katiebayes, line 47: with the unknown fields and how many of them being unknown, I like the spread operator to just grab them all as a single object to propagate. To keep consistency with other parsers to return no customContext when it isn't present (rather than `customContext = {}` or `customContext: undefined`), the const itself should be undefined unless needed, so I'm assigning the `rest` object here only if it contains anything.

Spread operators fallback to empty objects which is usually great.

lib/propagation/aws.js
```diff
@@ -20,41 +20,60 @@ function marshalTraceContextv1(context) {
...
-      customContext = Object.assign({}, customContext, {
-        [name]: value,
-      });
-    }
+function unmarshalTraceContext(header = "") {
+  let parentSpanId, customContext;
+  const supportedFields = ["root", "self", "parent"];
+
+  // parse out each field to key/value pairs and reduce to single object
+  let { root, parent, self, ...rest } = header
```
katiebayes, line 34: Using a deconstruct and map/reduce here to kind of collapse the step of turning the string into an object. This way it's a little easier to incorporate the case insensitivity and it's a bit clearer/more semantic below for the self/parent/root handling.

lib/propagation/aws.js
```diff
@@ -20,41 +20,60 @@ function marshalTraceContextv1(context) {
...
+    .split(";")
+    .map(each => {
+      let [key, value] = each.split("=");
+      // return lowercase keys for parsable fields to allow case insensitivity
+      if (supportedFields.includes(key.toLowerCase())) {
+        return { [key.toLowerCase()]: value };
```
katiebayes, line 40: As noted above, only lowercasing specific keys that won't be directly propagated

katiebayes:
I'll catch up with folks in the am on where to go with this. I have this change (Parent -> Self) also ready to go on another branch, so I can merge that in when we confirm. All this is/will continue to be documented in the spec so we can eval python and ruby on the same lines: https://github.com/katiebayes/beeline-nodejs/compare/fix-aws-hooks...revert-to-self

paulosman:
> [edit2] Ok, thanks to @paulosman continuing to test how the ALB responds to things it's become clear that the `svcA` -> `ALB` -> `svcB` scenario actually doesn't work because `svcA` will have a hex string as the trace ID and the ALB will refuse to use that as a trace ID, needing something in the form of `1-version-traceid`. We're going to need to change how all of our parsing of the AWS header works in all beelines and `honeyalb`. This exceeds the scope of this portion of the project, which is to focus on providing hooks for users to choose their own header parsing, along with a few sensible defaults. I support not caring about how we parse AWS headers right now because it's all going to change soon, so let's land this and move on with our work. We'll come back and fix it up later.

Scratch this. I was sending the wrong header value. ALBs _will_ respect a Parent segment of the `X-Amzn-Trace-Id` header, if present. This means the 2 scenarios you outlined will both work if a beeline sends its span ID as the `Parent` segment in the outgoing header. That will be picked up by the load balancer which will log the value of the Root segment as the trace ID and the `Parent` segment as the parent ID. The span ID will be a newly generated ID (in the `1-version-traceid` format) which will be propagated as `Self` in the header received by `svcB`.

lib/propagation/aws.js
```diff
@@ -20,40 +22,75 @@ function marshalTraceContextv1(context) {
...
+  // check for aws format, add if not there
+  // AWS trace fields should have a 1-1234567890-[id] format
+  // [version]-[unix timestamp in seconds]-[id]
+  const regex = /^1-([\d]||[a-f]){10}-/g;
+
+  // if either of these ids are not in AWS format, convert them before serializing the header
+  // Span.payload.startTime is stored in ms, so convert to seconds
+  if (!traceId.match(regex)) {
+    const rootSpanTimestamp = Math.floor(util.rootSpan(context).startTime / 1000);
+    traceId = `${VERSION}-${rootSpanTimestamp}-${traceId}`;
```
katiebayes, line 34: @paulosman what are your thoughts on this? I've added some stuff here to grab the Span's timestamp, convert to seconds and add it in here to mimic aws headers fully. I'm not sure if this is needed or helpful anymore for our use case.

paulosman: @katiebayes I don't think so, let's remove it.

katiebayes:
Discussions have happened, decisions have been made, this should be in line with them now! Re-requesting review if anyone has a moment

lib/propagation/propagation.test.js
```diff
@@ -40,7 +65,21 @@ cases(
...
+        parentSpanId: "0102030405",
+        customContext: {
+          userID: "1",
+          errorMsg: "failed to sign on",
+          toRetry: "true",
+        },
+      },
+    },
+    {
+      name: "lowercase keys work, custom keys remain unchanged",
```
katiebayes, line 80: new test for case insensitivity

lib/propagation/propagation.test.js
```diff
@@ -72,6 +111,32 @@ cases(
...
+        parentSpanId: "1-abcdef123456",
+        customContext: {
+          userID: "1",
+          errorMsg: "failed to sign on",
+          toRetry: "true",
+        },
+      },
+    },
+    {
+      name: "aws header with empty string parent",
```
katiebayes, line 128: added tests for the empty string, which should be treated the same as undefined. Empty string root fails to parse, empty string parent/self uses root as parent span id.

lib/propagation/util.js
```diff
@@ -7,3 +7,17 @@ exports.currentSpanId = currentSpanId;
 function currentSpanId(context) {
   return context.stack[context.stack.length - 1].payload[schema.TRACE_SPAN_ID];
 }
+
+exports.currentSpan = currentSpan;
```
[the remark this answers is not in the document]
katiebayes: Nope that was lingering from another implementation approach sorry. Removing...

lib/propagation/aws.js
```diff
@@ -20,40 +30,63 @@ function marshalTraceContextv1(context) {
     contextToSend = ";" + elements.join(";");
   }
 
-  return `Root=1-${traceId};Parent=${spanId}${contextToSend}`;
+  // if either of these ids do not already prefix with version, convert before serializing
+  if (!traceId.startsWith(`${VERSION}-`)) {
+    traceId = `${VERSION}-${traceId}`;
+  }
```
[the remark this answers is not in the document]
katiebayes: @toshok ok I'll just remove it

lib/propagation/aws.js
```diff
@@ -20,40 +30,63 @@ function marshalTraceContextv1(context) {
...
+    parentSpanId = self;
+  }
+  // if there is no "Self=" but there's a "Parent=",
+  // use "Parent=" as the parent span id
+  else if (definedAndNotEmpty(parent)) {
+    parentSpanId = parent;
+  }
+  // if no self or parent, use the "Root=" as the parent span id as well
+  else {
+    parentSpanId = root;
```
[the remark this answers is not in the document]
katiebayes: @toshok The map/reduce handles the parsing, and field assignment (and lowercasing), and this section handles the logic around the parsed fields, so I don't see this as uneven. `split` returns an Array, go through and split those, return the fields with a single object, and move on. When you say you'd prefer a loop, can you clarify specifically what you'd like to see here? Do you want to parse the fields and loop through the fields instead of the if statements?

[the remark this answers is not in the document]
katiebayes: the `||` statement here is a good call though, switching over

lib/instrumentation/trace-util.test.js
```diff
@@ -32,8 +32,9 @@ describe("getTraceContext", () => {
         },
       },
       {
-        name: "root / parent",
-        headerVal: "Root=1-5759e988-bd862e3fe1be46a994272793;Parent=53995c3f42cd8ad8",
+        name: "root / self / parent",
```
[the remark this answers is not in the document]
katiebayes: Added!

lib/propagation/aws.js
```diff
@@ -20,40 +30,63 @@ function marshalTraceContextv1(context) {
...
+  return `Root=${traceId};Parent=${spanId}${contextToSend}`;
 }
 
 // unmarshalTraceContext takes a string trace header and returns a context
 exports.unmarshalTraceContext = unmarshalTraceContext;
 
-function unmarshalTraceContext(header) {
-  let traceId, parentSpanId, customContext;
+function unmarshalTraceContext(header = "") {
+  let parentSpanId, customContext;
```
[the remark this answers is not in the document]
katiebayes: Done! We actually don't even need `let parentSpanId` with the change from above. It'll fall back to undefined regardless and no need for reassignment.

lib/propagation/propagation.test.js
```diff
@@ -25,13 +25,38 @@ describe("marshaling", () => {
...
 });
 
-describe("marshaling aws", () => {
-  test("version string prefix", () => {
-    expect(aws.marshalTraceContext(testContext)).toEqual(
-      "Root=1-abcdef123456;Parent=0102030405;userID=1;errorMsg=failed to sign on;toRetry=true"
-    );
-  });
-});
+cases("marshaling aws", opts => expect(aws.marshalTraceContext(testContext)).toEqual(opts.header), [
```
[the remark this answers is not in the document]
katiebayes: Done. Also created `w3c.test.js`

lib/propagation/aws.js
katiebayes: Totally. I'll make a note to look at that more in-depth once the stuff currently in flight gets out.

lib/propagation/aws.js
```diff
@@ -20,40 +21,46 @@ function marshalTraceContextv1(context) {
     contextToSend = ";" + elements.join(";");
   }
 
-  return `Root=1-${traceId};Parent=${spanId}${contextToSend}`;
+  return `Root=${traceId};Parent=${spanId}${contextToSend}`;
```
toshok, line 24: 🤨  hrm.  really not sure what to do here.  we need to guarantee that a `1-` gets prepended by this marshaler (for traces that begin in this service), but that means we need to remove it in the unmarshal function below.

does the go beeline currently remove it when unmarshaling?  the trace ID isn't supposed to include the version, is it?

lib/instrumentation/trace-util.test.js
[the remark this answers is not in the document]
toshok: where?

lib/propagation/aws.test.js
```diff
@@ -0,0 +1,168 @@
...
+        dataset: "testDataset",
+        stack: [new Span({ [schema.TRACE_SPAN_ID]: "0102030405" })],
+        traceContext: {
+          userID: 1,
+          errorMsg: "failed to sign on",
+          toRetry: true,
+        },
+      },
+      header:
+        "Root=abcdef123456;Parent=0102030405;userID=1;errorMsg=failed to sign on;toRetry=true",
```
toshok, line 38: these are invalid headers, aren't they?  supposed to be `Root=1-${traceId}`

lib/propagation/aws.js
paulosman: The Go beeline just passes along a traceID (no pretending of version number). I'm wary of trying too hard to guarantee the format of the ID otherwise we'd have to mirror in unmarshal and we should be including timestamps, etc. Through experimentation, the ALB will grab a root that doesn't start with the version.

lib/propagation/aws.test.js
paulosman: See above - doesn't matter to the load balancer.

lib/propagation/aws.js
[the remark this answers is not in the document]
toshok: I'd like to see `.reduce()` removed.  I understand what it's doing, but that textual description doesn't help future code readers (nor will c&p'ing it into a comment, I think.)

lib/instrumentation/trace-util.test.js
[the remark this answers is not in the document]
toshok: where?  I didn't see it.

lib/propagation/aws.test.js
toshok: right, but again with the interop - will other tracing systems bail when they see that header?

katiebayes: These tests use both prefixed and non to ensure that headers pass through unchanged regardless of prefixing

lib/propagation/aws.js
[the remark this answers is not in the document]
katiebayes: @toshok .reduce is just combining the logic from map into a single object. In the interest of avoiding more back-and-forth on this since it's holding up interop config, would you mind either pushing a commit with your change or outlining exactly what you'd like this to be?

lib/propagation/aws.test.js
toshok: that just means we're testing for (imo) broken behavior - that's not the point.

lib/instrumentation/trace-util.test.js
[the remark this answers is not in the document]
katiebayes: Oh I had realized before that we weren't testing parent at all, so incorporated a bunch. I just also added a test that is _only_ root and parent and no self.

lib/propagation/aws.js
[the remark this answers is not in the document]
toshok: I'm not asking for an explanation - I know what it's doing.  I also don't disagree that it's a familiar pattern.  I think it's a _bad_ pattern.

I feel like I've been quite explicit.  I'd like to see the reduce call gone.  either a loop or a `.forEach` (instead of _both_ `.map` and `.reduce`) that builds up the object we want.

lib/propagation/aws.test.js
katiebayes: Broken in what way? We don't remove prefixes when they come in, and we don't add them when they propagate. Those should be tested, shouldn't they? If the concern is around interop with w3c, aws already does not use w3c-standard ids, even if we were to strip them out.

katiebayes: I hope and believe that aws will likely update that as opentelemetry continues to gain traction

lib/propagation/aws.js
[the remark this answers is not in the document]
katiebayes: @toshok do you want to maybe jump on and rewrite this together irt?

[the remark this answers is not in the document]
katiebayes: Changing the map.reduce will also change the logic below, and will still need the case insensitivity introduced if it's based on the previous version. I'm happy to take a stab, but I'd anticipate some changes, so might just make more sense and save some time to just go through this together :) I'm also totally fine if you'd rather write the function.

lib/propagation/aws.test.js
toshok: the concern is interop with other things using aws headers.  w3c/otel don't matter.

dsoo: Thinking about the goal here - our goal is primarily to propagate traces in a way that creates legitimate traces when someone is using Honeycomb. A secondary goal would be to fail in the face of invalid trace propagation data that allows for the easy identification of the error.

What are the potential options upon receiving an invalid trace header?
* Reject the header entirely, don't propagate the trace. Right now this will result in a split trace, two different traceIDs representing the trace. If we throw some sort of warning into the newly created trace, this could help make it easier to debug the problem. e.g. `propagation_error: Invalid trace header: <header>` If this is a standardized header this could be super-cool and we could even automatically flag in the UI that there is invalid trace propagation happening somewhere.
* Ignore the fact that the header is invalid and continue propagation. This will possibly result in something further down that is propagated to error out, causing the rest of the trace to be missing or split. It may be a bit harder to debug, as it will point to the beeline and not to the original source of the invalid header.
* Repair the invalid header, continue propagation. This may result in a correct trace, or it may not, depending on how we handle the source trace.

My gut says the first choice is probably best in terms of transparency and debuggability, if we know for certain that a header of this form is invalid. If someone wants to repair invalid trace headers they can always create a custom hook to do so.

katiebayes:
Ok, after going over some of the active questions with the team, it seems like there are still some questions around AWS interoperability with honeycomb and w3c. Our main goal is to get support out there for w3c interop, so I'm going to close this for now and the upcoming config PR will not include AWS hooks. This will be the same in other beelines as well.

AWS will happen, but there are too many open questions to get it out at the same time as w3c. We will come back to it!

PR #233, opened 2020-07-28
[draft] Enable trace hook configs for interop

katiebayes:
**This is back in draft re: some beeline design discussion, I can't reclassify it as a draft but please consider it in-progess**
** ** note: this is also holding on #231 ** **

I don't believe this introduces breaking changes; trace parsing will fall back to the previously existing behavior if hooks are not set in the configuration.

- New configuration settings `httpTraceParserHook` and `httpTracePropagationHook` will allow users to pass functions in for trace propagation (from an active trace context) and trace parsing (from a request header)
- pre-built hooks in the API for aws and w3c parsing and propagation to streamline the process of switching over. These hooks can be passed through as config settings. One step more "magical" than the unmarshal/marshal functions, these hooks will handle finding and setting the trace header keys, and for w3c will handle parsing and propagation of both `tracestate`(where applicable) and `traceparent`
- for backwards compatibility, if the new configs are not set, the existing trace util with header detection (switch statement) will be used
- also keeping the existing traceIdSource setting and the most of the existing scope of config options (in the LibhoneyEventAPI class)

#### using the pre-built hooks
require the beeline as usual and pass the hook straight to the config. This example will propagate w3c-standard headers on outbound requests
```

beeline({
    writeKey: process.env.HONEYCOMB_WRITE_KEY,
    dataset: process.env.HONEYCOMB_DATASET,
});
```

#### Custom hooks using unmarshal functions
Users can pull down the unmarshal/marshal (or the hook if they really want to) and pass a custom function to the beeline for parsing or propagation. This example will parse a w3c header if detected, otherwise will default to honeycomb.

```

beeline({
    writeKey: process.env.HONEYCOMB_WRITE_KEY,
    dataset: process.env.HONEYCOMB_DATASET,
    httpTraceParserHook: customParserHook,
});

// custom parser must return beeline context format
// using unmarshal functions will handle that
function customParserHook(req) {
    const { headers } = req;

    // header keys are exposed in the API for this kind of use  
    const [ w3cHeader, honeycombHeader ] = [ beeline.w3c.TRACE_HTTP_HEADER, beeline.honeycomb.TRACE_HTTP_HEADER ];

    if (headers.w3cHeader) {
        return beeline.w3c.unmarshalTraceContext(headers.w3cHeader);
    }

    return beeline.honeycomb.unmarshalTraceContext(headers.honeycombHeader);
}
```

**to do**
- [x] update API.md doc
- [ ] add tests for the hook config settings

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -92,6 +92,8 @@ let rootSpan = beeline.startTrace({
 });
 ```
 
+---
+
 #### finishTrace()
 
 ```javascript
@@ -112,6 +114,8 @@ fs.writeFile(filePath, fileContents, err => {
   beeline.finishTrace(rootSpan);
 });
 ```
+
+---
 
 #### withTrace()
 
@@ -146,9 +150,13 @@ console.log(
 );
 ```
 
+---
+
 #### getTraceContext()
 
 Returns the current active trace, which is the needed input for any [marshalTraceContext](#marshaltracecontext) function.
+
+---
 
 #### startSpan()
 
@@ -174,6 +182,8 @@ fs.writeFile(filePath, fileContents, err => {
   beeline.finishSpan(span);
 });
 ```
+
+---
 
 #### startAsyncSpan
 
@@ -201,6 +211,8 @@ beeline.startAsyncSpan({
    }
 });
 ```
+
+---
 
 #### finishSpan()
 
@@ -240,6 +252,8 @@ beeline.finishSpan(parentSpan);
 // parentSpan's data has been sent to honeycomb.
 ```
 
+---
+
 #### withSpan()
 
 If you're doing something synchronously (looping, for instance, or using a synchronous node api) you can use `withSpan` to wrap this operation. It safely wraps the invocation of `fn` with `startSpan()` and `finishSpan()` calls. It returns the return value of fn, so can be used in a expression context.
@@ -262,6 +276,8 @@ let sum = beeline.withSpan(
   }
 );
 ```
+
+---
 
 ### Interprocess trace propagation
 
@@ -295,6 +311,8 @@ let traceContext = beeline.marshalTraceContext(beeline.getTraceContext());
 console.log(traceContext); // => 1;trace_id=weofijwoeifj,parent_id=owefjoweifj,context=SGVsbG8gV29ybGQ=
 ```
 
+---
+
 #### `unmarshalTraceContext()`
 
 ```javascript
@@ -320,6 +338,8 @@ let { traceId, parentSpanId, dataset, customContext } = beeline.unmarshalTraceCo
 
 let trace = startTrace({ name }, traceId, parentSpanId, dataset, customContext);
 ```
+
+---
 
 ### Trace interoperability
 
@@ -451,6 +471,8 @@ let trace = startTrace({ name }, traceId, parentSpanId, dataset, customContext);
     });
     ```
 
+---
+
 #### `TRACE_HTTP_HEADER`
 
 The HTTP header that the beeline uses both for sending and receiving trace context. The value is `"X-Honeycomb-Trace"`.
@@ -507,19 +529,27 @@ service2.on("something", async payload => {
 });
 ```
 
+---
+
 #### `w3c.TRACE_HTTP_HEADER`
 
 Contains the w3c trace header key expected for w3c parsing and propagation. Can be used in custom trace parse or trace propagation hooks. [Example above](#w3cmarshalTraceContext)
 
+---
+
 #### `aws.TRACE_HTTP_HEADER`
 
 Contains the aws trace header key expected for aws parsing and propagation. Can be used in custom trace parse or trace propagation hooks. [Example above](#awsmarshalTraceContext)
+
+---
 
 ### Adding context to spans
 
 _[TODO: This part is most likely to see changes as we figure out inter-process trace context propagation]_
 
 There are two axes to useful traces: depth and width. Depth is a measure of how many operations are performed, and width is a measure of how much information there is to associate with each operation. Depth is addressed by the startSpan/finishSpan. Width is addressed by the methods in this section.
+
+---
 
 #### span.addContext()
 
@@ -530,6 +560,8 @@ span.addContext(contextMap);
 
 Adds all key/value pairs in `contextMap` as toplevel fields on the current span. Context added with this method is _only_ attached to the current span. For attaching context to all spans sent after this call, use `addTraceContext`. Unlike `addContext` below, `span.addContext` does not prepend `app.`. It is primarily intended for use in instrumentations that maintain their own prefixes.
 
+---
+
 #### addTraceContext()
 
 ```javascript
@@ -537,6 +569,8 @@ beeline.addTraceContext(contextMap);
 ```
 
 Adds all key/value pairs in `contextMap` as toplevel fields on the current span, prepending `app.` This context will be attached to all spans sent after the call to `addTraceContext`, and all fields will be propagated to downstream/outbound http/https requests.
+
+---
 
 #### addContext()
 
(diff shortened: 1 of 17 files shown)


lib/propagation/w3c.js
```diff
@@ -63,3 +67,51 @@ function marshalTraceContextv1(context) {
   // we will revisit as opentelemetry sampling evolves
   return `${VERSION}-${context.id}-${spanId}-01`;
 }
+
+exports.httpTraceParserHook = httpTraceParserHook;
+
+// parserHook functions do not require users to pass an input
+// the instrumentation will provide the request object
+function httpTraceParserHook(req = {}) {
```
katiebayes, line 75: it seemed prudent to expose hooks for w3c that will handle the `tracestate` and `traceparent` headers, so the hook can be passed as-is. Users can pull this down and pass it along as a config setting to get up and running quickly, or they can pass a custom http hook for more specific customization needs, or built-in logic around which header to prioritize

lib/api/index.js
```diff
@@ -141,6 +146,10 @@ module.exports = {
     }
   },
 
+  validateOpts(opts) {
```
katiebayes, line 149: I've added this to the api for use in various instrumentations; user-defined options seem to be passed directly to the constructor in the `LibhoneyEventAPI` (or mock) so are not in scope for instrumentations like express and fastify, that handle incoming requests. Since we need to trigger header parsing on incoming requests, but it seemed risky/unnecessary to alter which data actually flows through the instrumentation functions, this function will expose the user-set options to the traceUtil that can access the options via libhoney.js, without passing this data through the implementation itself.

katiebayes:
Still need to add some tests but I'm opening this up for review since there's plenty to look at :)

docs/API.md
```diff
@@ -13,13 +13,57 @@ const beeline = require("honeycomb-beeline")();
 /* beeline.traceActive() works, as do all the other calls below */
 ```
 
-There are a few surfaces to the API:
+Contents
```
katiebayes, line 16: This doc is getting pretty long, so it seemed like a good moment to add a TOC for quick scan-ability

lib/instrumentation/express.js
```diff
@@ -78,13 +78,10 @@ function wrapErrMiddleware(middleware, debugWrap) {
...
-  req,
-  res,
-  next
-) => {
-  // parse incoming trace headers into a span context
-  let spanContext = traceUtil.getTraceContext(traceIdSource, req);
+const getMagicMiddleware = opts => (req, res, next) => {
+  const { userContext, parentIdSource, packageVersion } = api.validateOpts(opts);
+
+  let spanContext = traceUtil.parseTraceHeader(req, api.validateOpts(opts));
```
toshok, line 84: this code is calling `api.validateOpts` twice.  one of those times it's passing the resulting object to `traceUtil.parseTraceHeader`, which is validating it _again_.

lib/instrumentation/express.js
```diff
@@ -177,46 +174,12 @@ const getMagicMiddleware = ({ userContext, traceIdSource, parentIdSource, packag
...
-    if (typeof opts.parentIdSource === "string" || typeof opts.traceIdSource === "function") {
-      parentIdSource = opts.parentIdSource;
-    } else {
-      debug(
-        "parentIdSource option must either be an string (the http header name) or a function returning the string request id"
-      );
-    }
-  }
-
-  let packageVersion = opts.packageVersion;
```
toshok, line 212: This validation reworking will result in the wrong config options being validated.  these options are not toplevel config options passed to the beeline configure method.  they're passed within the `express:` suboptions.  i.e. the user doesn't do:

```
  parentIdSource: "..."
});
```

they need to do:

```
  express: {
    parentIdSource: "..."
  }
});
```

most importantly here - `opts` in this method is _not_ the toplevel `opts` inside api/index.js.

if you check instrumentation.js, the `opts` in this method is supplied as:

```
        return instrumentLoad(mod, request, parent, opts[request] || {});
                                                              ^^^
```

lib/config.js
```diff
@@ -0,0 +1,32 @@
...
+    debugMsg:
+      "traceIdSource option must either be an string (the http header name) or a function returning the string request id",
+    allowedTypes: ["string", "function"],
+  },
+  {
+    name: "parentIdSource",
+    debugMsg:
+      "parentIdSource option must either be an string (the http header name) or a function returning the string request id",
+    allowedTypes: ["string", "function"],
+  },
```
toshok, line 31: I think these need to go away and revert back to instrumentation options.

lib/api/libhoney.js
```diff
@@ -84,6 +85,8 @@ module.exports = class LibhoneyEventAPI {
 
     this.defaultDataset = opts.dataset || process.env["HONEYCOMB_DATASET"] || defaultName;
 
+    this.opts = opts;
```
toshok, line 88: I don't think we should be storing `opts` on this instance.  we're already storing the options we care about (that affect this instance, like `defaultDataset` above).  what's to be gained by storing the entire object as well?

lib/api/index.js
toshok: adding it here means it's accessible from the outside.  I don't think we want/need that.  

Validation of config options should imo proceed top-down, with each layer validating its options at _initialization time_.  introducing a cycle where code calls back through api.validateOpts from way down deep feels very wrong - options specific to the toplevel api should have already been validated, and options specific to the layer way down deep should be validated closer to where they're used - that way down deep layer.

lib/config.js
```diff
@@ -0,0 +1,32 @@
+/* eslint-env node */
+
+exports.configOptions = [
```
toshok, line 3: I think this is the wrong split to make when it comes to validating config options.

imo the config options that are specific to `opts` passed to api/index.js's `configure` function should live in that file (maybe even that _function_.)  This file should export `validateOptions(options, configSpec) => object` and should contain the loop that validates things, returning the validated (or overridden to `undefined`) object.

lib/api/mock.js
```diff
@@ -3,14 +3,16 @@ const path = require("path"),
...
   Span = require("./span");
 
 module.exports = class MockEventAPI {
-  constructor(opts) {
+  constructor(opts = {}) {
     this.constructorArg = opts;
     this.traceContext = {};
     this.sentEvents = [];
     this.traceId = 0;
+    this.opts = opts;
```
toshok, line 15: again - why are we storing opts here?  it's already stored as `constructorArg`.

lib/instrumentation/express.js
katiebayes: Ah thanks for the clarification!

lib/api/mock.js
katiebayes: I think I need some clarification of the configuration options here if you could humor me: I like the idea of validating options once at configuration, but I was also using that to expose new and necessary configurations to the instrumentation. How would you go about exposing a general parser/propagation hook config setting to multiple instrumentations if not via the api?

katiebayes: I know python also has a situation of instrumenting for different frameworks, I'll find out what the plan is there

lib/api/index.js
katiebayes: I think I need some clarification of the configuration options here if you could humor me: I like the idea of validating options once at configuration, but I was also using that to expose new and necessary configurations to the instrumentation. How would you go about exposing a general parser/propagation hook config setting to multiple instrumentations if not via the api? Or does this mean that general trace interop configurations would not be available as general configuration at all?

katiebayes: Moving the comment here to continue this thread: I'll find out in the am what the plan is in python as well, since it's a pretty similar situation of very instrumentation-specific options

katiebayes:
Tests are failing on aws, I stopped work there until I get #231 in; will fix once that's merged!

katiebayes:
Closing this as well, will get something up first for custom hooks only

PR #241, opened 2020-08-05
validation improvements in w3c

katiebayes:
Per discussion earlier today, bringing this beeline in line with the w3c specifications for where we should have opinions on id validation. Most importantly, we shouldn't propagate any w3c headers that don't meet w3c specification. w3c validation boils down to character count and allowed characters; we agreed to use regex for this, similar to opentelemetry.

Since we're using the opentelemetry parse function, I'm leaving the parse version handling there (using a regex in ruby for this).

Spec for those curious: https://www.w3.org/TR/trace-context/#versioning-of-traceparent

### DIFF
--- a/lib/propagation/w3c.js
+++ b/lib/propagation/w3c.js
@@ -67,12 +67,14 @@ function marshalTraceContextv1(context) {
 
   // do not propagate non-standard ids
   if (!context.id.match(TRACE_ID_REGEX)) {
-    debug("unable to propagate w3c trace header: trace id must be a 32-character hex string");
+    debug(
+      "unable to propagate w3c trace header: trace id must be a 32-character hex encoded string"
+    );
     return "";
   }
 
   if (!spanId.match(SPAN_ID_REGEX)) {
-    debug("unable to propagate w3c span header: span id must be a 16-character hex string");
+    debug("unable to propagate w3c span header: span id must be a 16-character hex encoded string");
     return "";
   }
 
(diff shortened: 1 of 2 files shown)


lib/propagation/w3c.js
```diff
@@ -59,6 +62,11 @@ function marshalTraceContext(context) {
 function marshalTraceContextv1(context) {
   const spanId = util.currentSpanId(context);
 
+  // do not propagate non-standard ids
+  if (!context.id.match(TRACE_ID_REGEX) || !spanId.match(SPAN_ID_REGEX)) {
+    return;
```
toshok, line 67: hm, this doesn't an indication as to why it was rejected.

I'd imagine in golang if the return type is `string`, then "" would be returned in that code.  I think that'd be a good change here (and would make api docs easier to deal with) - undefined and "" are both falsey, but the latter means that folks don't have to actually _check_ if they don't want to.  anywhere they'd put a result from `marshalTraceContext`, they can still put it.

if we actually do want to do error handling there (in golang to continue the ... analogy), we would probably change the return type to `(string, error)` with these being:

1. `return nil, errors.New("illegal w3c trace id")`
2. `return nil, errors.New("illegal w3c span id")`

respectively.  That means instead of returning here we should _probably_ throw an exception.

I don't know what the error handling story is for the other beelines.  Honestly throwing an exception feels like the right thing for JS, but until we've decided more widely, "" feels like the right thing to do.

katiebayes: Agreed. Changing to empty string and adding some debug messaging as well

katiebayes:
Re-requesting (sorry), just want to make sure we're good on the debug usage here

katiebayes:
Confirmed via slack, we're good here!

PR #242, opened 2020-08-06
Custom hooks for parsing and propagation

katiebayes:
Let's try getting custom hooks configuration out there without the pre-built ones. I'm hoping I've found a good middle ground here.

**Updated Notes**
- Now, rather than try to keep the hooks in the exact same scope as other configurations, the current version passes the configuration straight through to propagation, where the hooks are filtered out and stored on the configuration flow (which is shimmed to module load)
- These hooks can then be accessed from propagation, directly by the traceUtil, i.e. directly by the instrumentation. (`traceUtil` stores functions that are shared by instrumentations and is in the same directory as the individual instrumentations)

**Notes:**
- The instrumentation team would like the custom hook configurations to be base configurations, similar to other hooks options, rather than have it nested under specific instrumentations
- ~~To keep the scope of the config options exactly the same, the custom hooks will now be attempted right in the LibhoneyEventAPI class; hooks and other config settings will not be returned anywhere else, just the return value of the custom hook~~ changed, see above
  - ~~This does mean error handling and debugging is being handled in that class as well, rather than in propagation or instrumentation files. I think this is fine.~~ changed, see above
  - Fallback/existing parsing and propagation will remain where it is
- Since it's going to be easier to break traces as users figure out their customized preferences, and since we have decided to avoid allowing the beeline to break applications, I've incorporated some try/catch patterns and added a whole lot of debug messaging indicating what's going on with tracing (particularly where applying custom hooks)
- We have made a team decision to use the "propagation context" (the object with traceId, spanId, customContext, and dataset) the standard and documented intermediary structure for both parsing and propagation. In other words, parserHooks must return a propagation context (this is unchanged), and unmarshal functions and propagation hooks must expect a propagation context as input (this is new in this beeline).
  - I have kept the execution field paths as a fallback for backwards compatibility, prioritizing the new expected structure


lib/api/libhoney.js
```diff
@@ -84,6 +84,11 @@ module.exports = class LibhoneyEventAPI {
 
     this.defaultDataset = opts.dataset || process.env["HONEYCOMB_DATASET"] || defaultName;
 
+    this.propagationHooks = {
```
katiebayes, line 87: Just making these exact two options available in the below method

lib/api/libhoney.js
```diff
@@ -102,8 +107,58 @@ module.exports = class LibhoneyEventAPI {
...
   }
 
+  customParseTraceHeader(req) {
+    const { httpTraceParserHook } = this.propagationHooks;
+    if (!httpTraceParserHook) {
+      // if no custom parser hook, bail to use defaults
+      return;
+    }
+
+    try {
```
katiebayes, line 117: wrapping this in a `try` so we can catch and log any exceptions. Seemed appropriate since this is a totally custom function

lib/api/libhoney.js
```diff
@@ -102,8 +107,58 @@ module.exports = class LibhoneyEventAPI {
...
+      }
+
+      // if a custom parser hook returns something truthy but not an object
+      // assume configuration error, user returned wrong type
+      if (typeof customParsed !== "object") {
+        throw new Error("httpTraceParserHook must return an object");
+      }
+
+      // check for keys in the parser hook that are not usable
+      const contextKeys = Object.keys(customParsed);
```
katiebayes, line 132: Just being extra thorough here, if I did some fancy custom stuff and accidentally returned `context.spanId` instead of `context.parentSpanId` I'd want to be able to find that quickly

katiebayes: I didn't intend to break on unsupported fields though, just log them on debugging. Handling missing trace ids and such is already handled elsewhere in the beeline.

lib/instrumentation/trace-util.js
```diff
@@ -31,6 +31,33 @@ function _includeSource(context, key) {
   });
 }
 
+exports.parseTraceHeader = (traceIdSource, req) => {
```
katiebayes, line 34: the functions in libhoney.js will return undefined if there is no custom hook in the configuration. This split allows the config option to stay in the libhoney object and never have to pass here (returning the configuration from those function and applying it here was called out as a concern in the previous PR)

lib/propagation/w3c.js
```diff
@@ -26,32 +23,26 @@ exports.unmarshalTraceContext = unmarshalTraceContext;
 // isRemote (optional)       -->  context.isRemote (nested field in context: {})
 // traceState (optional)     -->  context.traceState (nested field in context: {})
 
-function unmarshalTraceContext(header) {
```
katiebayes, line 29: Per w3c spec, we actually should not be checking strictly against version here. Removing this completely and letting the otel parser handle it

lib/propagation/w3c.js
```diff
@@ -63,22 +54,25 @@ function marshalTraceContext(context) {
...
-  if (!spanId.match(SPAN_ID_REGEX)) {
+  if (!parentSpanId.match(SPAN_ID_REGEX)) {
     debug("unable to propagate w3c span header: span id must be a 16-character hex encoded string");
     return "";
   }
 
   // currently we do not propagate traceFlags
   // we will revisit as opentelemetry sampling evolves
-  return `${VERSION}-${context.id}-${spanId}-01`;
+  return `00-${traceId}-${parentSpanId}-01`;
```
katiebayes, line 77: Just hardcoding this now, there's no reason to store the version elsewhere because we don't need to check against it. Doing the same thing in ruby

lib/api/index.js
```diff
@@ -51,6 +51,18 @@ module.exports = {
     return tracker.getTracked();
   },
 
+  customParseTraceHeader(req) {
```
katiebayes, line 54: These use slightly different naming because they _only_ return the results of custom hooks. Built-in treatment of parsing and propagation is in the trace_util file

lib/api/index.js
```diff
@@ -51,6 +51,18 @@ module.exports = {
...
+    return apiImpl.customParseTraceHeader(req);
+  },
+
+  getPropagationContext() {
```
katiebayes, line 58: To make it easy for users to access the propagation context format of the current trace when writing custom hooks, I'm exposing this function in the api. It just fetches the current active trace and returns it in propagation context format.

lib/api/libhoney.js
```diff
@@ -102,8 +107,81 @@ module.exports = class LibhoneyEventAPI {
...
+
+      if (!customParsed) {
+        // bail if no return from custom hook
+        return;
+      }
+
+      // if a custom parser hook returns something truthy but not an object
+      // assume configuration error, user returned wrong type
+      if (typeof customParsed !== "object") {
+        throw new Error("httpTraceParserHook must return an object");
```
MikeGoldsmith, line 128: Instead of throwing an error to the catch statement that just logs and eats it, I think it would be more clear to log here and return.

lib/api/libhoney.js
```diff
@@ -102,8 +107,81 @@ module.exports = class LibhoneyEventAPI {
...
+
+      if (!customSerialized) {
+        // bail if no return from custom hook
+        return;
+      }
+
+      // if a custom propagation hook returns something truthy but not an object
+      // assume configuration error, user returned wrong type
+      if (customSerialized && typeof customSerialized !== "object") {
+        throw new Error(
```
MikeGoldsmith, line 167: As as above, I think it would be better to log and return directly here instead of throwing an error that just gets eaten.

MikeGoldsmith (dismissed):
LGTM - just a couple of suggestions about not throwing errors when they are just eaten inside a try/catch.

lib/propagation/util.js
```diff
@@ -7,3 +7,19 @@ exports.currentSpanId = currentSpanId;
 function currentSpanId(context) {
   return context.stack[context.stack.length - 1].payload[schema.TRACE_SPAN_ID];
 }
+
+exports.getPropagationContext = getPropagationContext;
+
+// getPropagationContext accepts the current execution context async hook
```
toshok, line 13:
```suggestion
// getPropagationContext accepts the current execution context
```

async_hooks is the facility we use to transmit the context around.

lib/propagation/w3c.js
toshok: doesn't the otel package have support for generating the header?  feels kinda weird to completely defer to otel when parsing but not when generating it.

lib/api/index.js
toshok: I don't think this belongs in the exported api.  why export this at the toplevel when any calling code is going to need to fall back?

lib/propagation/w3c.js
MikeGoldsmith: It does, but requires you to use the OTel HttpTextPropagator implementation instead of exposing a utility function like for `parseTraceParent`.
https://github.com/open-telemetry/opentelemetry-js/blob/master/packages/opentelemetry-core/src/context/propagation/HttpTraceContext.ts#L66

lib/api/libhoney.js
toshok: I'd say:

1) add a `configure(opts)` in propagation/index.js (ala the one in instrumentation.js)
2) call that method in the toplevel configure method (in lib/index.js).  the code in `propagation/index.js` takes care to deal with the propagation hooks, so we can remove this.
3) the `custom` methods below become.. not custom and just call through to the appropriate methods in propagation/index.js, which do both the custom and fallback behavior.  trying not to bikeshed the names more than "they shouldn't have 'custom' in them" at the moment.
4) move the resulting method bodies up to the toplevel api - there's nothing libhoney-specific about them.

This should hopefully get rid of the circular API calls.  `propagation/` code has no real reason to call through the api at all, and adding things to the toplevel api only to serve our own `propagation/` internals means we're stuck supporting more things.

lib/api/libhoney.js
```diff
@@ -102,8 +107,81 @@ module.exports = class LibhoneyEventAPI {
...
+
+      return customParsed;
+    } catch (error) {
+      debug(`unable to apply httpTraceParserHook: ${error}`);
+      return;
+    }
+  }
+
+  // to keep options scope in this object, try to serialize with custom hook here
+  // return undefined if no custom hook
```
toshok, line 149: this goal is leading to a bad API, imo.  we should not be keeping config scope so high.

lib/propagation/w3c.js
toshok: ahhh, that's really too bad.  not as pretty, but if the spanContext type is less likely to rev than the serialization format, could be something like:

```
  let rv;
  // spanContext below is the otel span context.  presumably a documented type, but maybe not?
  otelPropagator.inject(spanContext, undefined, (_, _, value) => rv = value);
  return rv;
```

lib/api/libhoney.js
katiebayes: Thanks, just some clarifying questions to try to avoid writing this too many more times :) :
1) what do you think the propagation/index.js function that receives the options should return? Can it just return the two hooks?
2) Can you say specifically what you mean by `takes care to deal with the propagation hooks`, deal with in what way and with what result at that stage? Are you just referring to the fact that it doesn't return all the options, just the hooks, or are you expecting it to do more?
3) And just to make sure, are we saying we're ok with applying the custom hooks directly in the top-level api, or are we expecting that to happen elsewhere?
4) If we're moving the resulting (non-"custom") method bodies to the toplevel api, but we don't want the propagation code to call the api anymore, where are we calling them and how do you envision passing the custom hook (or result of it) to the instrumentations that handle incoming calls? Trace-util could probably pull in the propagation directly instead of via the api.

Thanks for clearing this up! When I have a clearer picture of the end result you're looking for I'm willing to take another stab.

lib/propagation/w3c.js
MikeGoldsmith: I think for now it's fair for us to use the string literal like we have above and avoid shoe-horning in OTel's W3C `inject` func. We don't really care about the version prefix or trace flags suffix right now, so just passing the in the traceId and parentSpanId is good enough.

lib/propagation/index.js
```diff
@@ -28,6 +29,32 @@ function marshalTraceContext(context) {
...
+    getHooks,
+  };
+})();
+
+// exported to run on require and module load with other configurations
+exports.configure = configure.setHooks;
+
+// expose custom hooks to instrumentation
+exports.customHooks = configure.getHooks;
+
```
katiebayes, line 56: As noted above and in an irt review, configs now pass directly into propagation, bypassing existing dependency flows completely. `setHooks` will run on the initial require along with other config hooks, and `getHooks` will return the hook that was set. An object seemed like it would be overkill here since we just need to set some variables and make them available elsewhere, so I opted for an anonymous function where the hooks will be stored on require and then can be fetched later from the same wrapped function.

katiebayes:
I'm going to add a few more tests today but this is reconfigured based on combined feedback and is fully functional so I'm opening it up for review!

lib/instrumentation/trace-util.js
katiebayes: Update here: Now this function can check for a custom hook before attempting to apply it. I like keeping the function that attempts to apply that config as a separate task rather than cluttering up this one, but the function gets called here. This is nice because we can succinctly access the hook and the function that applies it, do the work here, and apply the default if it's either not set or fails to apply.

lib/propagation/honeycomb.js
```diff
@@ -49,16 +51,22 @@ function marshalTraceContextv1(context) {
...
+    switch (version) {
+      case "1":
+        return unmarshalTraceContextv1(payload);
+    }
+    debug(`unrecognized trace context version ${version}.  ignoring.`);
+    return;
+  } catch (error) {
+    debug(
+      `unable to parse trace header: expected string value of "traceparent", received ${header}`
+    );
```
toshok, line 68: this looks wrong - at best the wrong message.  this looks c&p from w3c.js.  should probably also include `${error}` in the string (same comment applies to w3c.js).

lib/propagation/util.js
```diff
@@ -1,9 +1,99 @@
...
+  // standard propagation context fields and the path to its value
+  return {
+    traceId: context.id,
+    parentSpanId: currentSpanId(context),
+    dataset: context.dataset,
+    customContext: context.traceContext,
+  };
+}
+
+exports.applyCustomParser = applyCustomParser;
```
toshok, line 30: imo (and pretty minor) these next two functions here really warrant their own file, as opposed to being added to the grab bag util.js (maybe hooks.js?)

at any rate, _this_ is where the configure function should live - so that the local bindings for `httpTraceParserHook` and  `httpTracePropagationHook` are accessible.  That way getHooks/customHooks can get dropped, and the code in trace-utils.js simplified.

toshok (changes requested):
this is close.  remaining comments are I think just 2 types:

1. copy problems in debug() calls
2. api split ending up with more code on calling side than we need (which requires more api as well)

lib/instrumentation/trace-util.js
```diff
@@ -31,6 +32,43 @@ function _includeSource(context, key) {
...
+  // check for a custom parser hook config setting
+  const { httpTraceParserHook } = propagation.customHooks();
+
+  if (httpTraceParserHook) {
+    // if a custom parser was set at configuration, try to apply that first
+    const customParsed = propagation.util.applyCustomParser(req, httpTraceParserHook);
+    if (customParsed) {
+      return customParsed;
+    }
+  }
```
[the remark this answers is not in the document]
toshok: peeling things back a bit further, and it doesn't have to be in this PR, but the body of `exports.getTraceContext` could be copied over to `propagation` (into a private method), the fallback behavior contained there, and then the above function could go away entirely.  Instrumentation code could call into `propagation.parseFromRequest` (likely through the toplevel beeline API).

lib/api/libhoney.js
toshok: sorry, just coming back to this now:

1. The configure functions don't return anything - they modify local state for that collection of code.
2. I meant that the code inside of propagation/ is the only piece that needs to know about the hooks.  nowhere else in the beeline does the code need to know about them, that I can tell..
3. not sure about this.
4. we don't want propagation/ code to call into the api, correct.  we _do_ want instrumentation/ code to call into the API, though.  trace-utils.js is really a collection of things that we should figure out how to get rid of/move to the toplevel API - it's a symptom that it's impossible to write a toplevel instrumentation without access to internals.

lib/propagation/honeycomb.js
katiebayes: yeah definitely a typo, fixed

lib/instrumentation/trace-util.js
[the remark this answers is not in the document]
katiebayes: Done, renamed, moved to a "propagation/hooks.js" file. I left the hook argument for the sake of portability, falling back to the configuration. This will make it easier for us to pass in hooks or move it in the future, if that's ok?

lib/propagation/util.js
katiebayes: moved to a propagation/hooks.js file

lib/propagation/index.js
```diff
@@ -28,6 +29,33 @@ function marshalTraceContext(context) {
   return honeycomb.marshalTraceContext(context);
 }
 
+const configure = (function() {
```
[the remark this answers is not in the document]
katiebayes: I don't have a problem with this per se, but global vars are generally considered not great in node, aren't they? Making the change because I'm not interested in digging in on this, but just wanted to put out there the get/set pattern was to keep in line with general best practices for variable scope.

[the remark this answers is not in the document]
katiebayes: Change made

lib/instrumentation/trace-util.js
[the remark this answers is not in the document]
toshok: why would it make it easier?  I'm not sure I understand the possible future use-case.  It's just as easy to change the body of parseFromRequest.

lib/propagation/hooks.js
```diff
@@ -15,14 +15,9 @@ function configure(opts) {
 
 exports.parseFromRequest = parseFromRequest;
 
-function parseFromRequest(req, parserHook = httpTraceParserHook) {
-  if (!parserHook) {
-    // if no custom parser hook, bail to use defaults
-    return;
-  }
```
toshok: didn't notice this before, but losing this check (and the one from `headersFromContext`) will throw an exception + output a debug() message.  neither of which is terrible, but if the default case remains "user doesn't configure hooks", it's going to be pretty noisy.

toshok (approved):

lib/propagation/hooks.js
katiebayes: Fair point. I'll re-add this check based on the now-locally-scope hook.

dsoo (approved):
👍 Looks good. My only nit is that it might be nice to make some variable names `propagationContext` instead of `context` just because context is so overloaded, but I'm not super strongly opinionated about it.

PR #249, opened 2020-09-17
early return if the beeline has already been configured

toshok:
Repeated calls to the toplevel configure function can yield errors like this:

```
The following modules were required before honeycomb-beeline: express
These modules will not be instrumented.  Please ensure honeycomb-beeline is required first.
```

This is because the toplevel configure function calls all the sub-configure functions, including our `instrumentation/` configure which checks for already-loaded modules, causing the warning above.

The function is meant to be idempotent, so early return if we're already configured.  We do, however, need a warning in this case as well - if the beeline configure call is passed arguments we should let the user know they aren't going to affect anything.

so, after this change:

```
beeline({ ... });
beeline({ ... }); // => "Beeline is already configured.  Further calls are allowed but should not be passed configuration options (as they will be ignored)."
```

warning feels kinda wordy, but /shrug

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -3,10 +3,35 @@ const api = require("./api"),
   instrumentation = require("./instrumentation"),
   propagation = require("./propagation");
 
-function configure(opts = {}) {
+let configured;
+
+function checkIfConfigureNeeded(opts) {
+  if (!configured) {
+    return true;
+  }
+
+  if (opts) {
+    console.warn(
+      "Beeline is already configured.  Further calls are allowed but should not be passed configuration options (as they will be ignored)."
+    );
+  }
+
+  return false;
+}
+
+function configureBeeline(opts) {
+  configured = true;
   api.configure(opts);
   instrumentation.configure(opts);
   propagation.configure(opts);
+}
+
+function configure(opts) {
+  const needConfigure = checkIfConfigureNeeded(opts);
+
+  if (needConfigure) {
+    configureBeeline(opts);
+  }
 
   return configure;
 }
--- a/lib/index.test.js
+++ b/lib/index.test.js
@@ -13,7 +13,25 @@ test("the default export function returns itself, so the properties are still th
 });
 
 test("the function is idempotent.  call it as often as you want.", () => {
+  const oldConsole = global.console;
+  global.console = { warn: jest.fn() };
+
   const beeline = require(".");
+  beeline();
+  expect(console.warn).not.toBeCalled();
+
+  // but we'll warn if you call it again with options
   beeline({ disableInstrumentation: true });
-  beeline({ disableInstrumentation: true });
+  expect(console.warn).toBeCalled();
+  expect(console.warn.mock.calls[0][0]).toEqual(
+    expect.stringContaining("Beeline is already configured")
+  );
+
+  console.warn.mockClear();
+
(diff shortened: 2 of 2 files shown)


lib/index.js
```diff
@@ -2,9 +2,23 @@
 const api = require("./api"),
   instrumentation = require("./instrumentation");
 
-function configure(opts = {}) {
-  api.configure(opts);
-  instrumentation.configure(opts);
+let configured;
+
+function configure(opts) {
+  if (configured) {
```
katiebayes, line 8: It might just be me but I find the way this is nested and the optsToPass really hard to read. A couple suggestions:

- all of the configure functions have default parameters of `{}`, so the one here is extra anyway. I'd nix the ternary here and just pass the opts along, it will simplify things for the same result.
- Maybe consider an object for the top-level variable and store `configured` as a nested key? Future proofing in case we need to add more?
- I would split out the logic that does the checks for configured and opts to a separate function mainly for the sake of legibility. I'm awful at naming but something like an `alreadyConfigured()`
that just returns false if the beeline is not configured

Mocked it up really quick to illustrate my thinking, but take all of this as very loose suggestion, ignore if not helpful etc etc. Just trying to be helpful!

```
let beeline = {};

function configure(opts) {
  if (alreadyConfigured(opts)) {
    return configure;
  }

  api.configure(opts);
  instrumentation.configure(opts);
  propagation.configure(opts);

  return configure;
}

function alreadyConfigured(opts) {
  if (!beeline.configured) {
    // beeline.configured is false, set it to true and return false for alreadyConfigured
    beeline.configured = true;
    return false;
  }

  // beeline is already configured and opts is truthy, print the warning and return true for alreadyConfigured
  if (opts) {
    console.warn(
      "Beeline is already configured.  Further calls are allowed but should not be passed configuration options (as they will be ignored)."
    );
  }
  
  return true;
}
```

katiebayes (approved):
No blocking concerns, see comment for some suggestions that might help though.

lib/index.js
MikeGoldsmith: I don't mind the nested `if`s but agree the overloaded `optsToPass` could be more clear. I like your separation of checking whether the beeline has been initialised and actually doing the init into separate functions.

MikeGoldsmith (commented):
Overall the change looks good and I'm good with verbose warning message. I agree with Katie's comments regarding refactoring to be a little cleaner.

What is the use-case for calling configure without `opts`? The original function just defaulted omitted opts to an empty object to initialise with. What is the result of `configure` used for?

lib/index.js
toshok: hrm.. I can separate out the check but not the `configured = true` assignment.  not going to introduce a `beeline` object. 👍

toshok:
the assumption is that you call the function to initialize the beeline:

```
const beeline = require("honeycomb-beeline")({ configArgs: here });

beeline.startTrace(...);
```

once this has been done, you can use the api without making the call:

```

beeline.startSpan(...);
```

I think a combination of docs being unclear and folks instrumenting, then c&p the beeline require call means we end up with a lot of calls to the configure function.  I expect that most of the time, these are without args and harmless.

The beeline will pull the dataset/writekey from the environment, so there's no need to pass config params most of the time.  That's why it's usually empty.

PR #251, opened 2020-09-24
Add prebuilt interop hooks

katiebayes:
This should wrap up interop:

- Adds prebuilt interop propagation hooks so configuration can look like this:

```

beeline({
    httpTraceParserHook: beeline.w3c.httpTraceParserHook,
});
```
- includes a testing suite refactor from a previous closed PR
- in aws, includes a fallback to self rather than parent if parent is not present. (Note: This is in line with the current behavior in this beeline to prioritize `Parent`, which was under discussion)
- adds utility functions for converting an object to string for propagation, and a string to an object for ingestion. Join characters default to w3c/honeycomb (`,`, `=`) but can be overridden


paulosman:
Re-opened as #253

PR #252, opened 2020-09-24
Changelog, circle tweaks

dsoo:
* Updates to the CircleCI build process so that it more closely matches our "state of the art" process, including:
   * Matrix for node versions
   * Publish to github releases in addition to npm.
   * General cleanup
* Added `CHANGELOG.md`
* Instructions for using `npm version` to update the build number.
* Removed node 10.0.0 from our build matrix - not sure why we had that specific version.
* Added node 14 to our build matrix.

(Note: The build yellow because I changed the name of the CircleCI check from `build-libhoney` to `build`, I will change this in the github configuration as soon as the PR is signed off.)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 # YAML Anchors to reduce copypasta
 
 # This is necessary for job to run when a tag is created
-filters_all_tags: &filters_all_tags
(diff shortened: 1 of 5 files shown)


MikeGoldsmith (approved):
Looks great @dsoo.

PR #253, opened 2020-09-29
Add prebuilt interop propagation hooks

paulosman:
Adds prebuilt interop propagation hooks so configuration can look like this:

```

beeline({
    httpTraceParserHook: beeline.w3c.httpTraceParserHook,
});
```

This also includes a testing suite refactor from a previously closed PR.

For the AWS propagator, this includes a fallback to self (rather than parent) if parent is not present. This PR also includes utility functions for converting an object to a string for propagation, and a string to an object for ingestion. The join characters default to w3c/honeycomb (,, =) but can be overridden

### DIFF
--- a/lib/propagation/aws.js
+++ b/lib/propagation/aws.js
@@ -33,20 +33,22 @@ function unmarshalTraceContext(header) {
 
   const split = header.split(";");
   for (const s of split) {
-    const [nameUnaltered, value] = s.split("=");
-    const name = nameUnaltered.toLowerCase();
-
-    if (name === "root") {
-      traceId = value;
-    } else if (name === "parent") {
-      parentSpanId = value;
-    } else if (name === "self") {
-      self = value;
-    } else {
-      customContext = Object.assign({}, customContext, {
+    const [name, value] = s.split("=");
+    switch (name.toLowerCase()) {
+      case "root":
+        traceId = value;
+        break;
+      case "parent":
+        parentSpanId = value;
+        break;
+      case "self":
+        self = value;
+        break;
+      default:
         // propagate trace headers without case changes
-        [nameUnaltered]: value,
-      });
+        customContext = Object.assign({}, customContext, {
+          [name]: value,
+        });
     }
   }
 
--- a/lib/propagation/aws.js
+++ b/lib/propagation/aws.js
@@ -31,6 +31,12 @@ function unmarshalTraceContext(header) {
       traceId = value;
     } else if (name === "Parent") {
       parentSpanId = value;
+    } else if (name === "Self") {
+      // use Self as parentSpanId if parentSpanId is undefined
+      // Parent will overwrite this value regardless of order
+      if (!parentSpanId) {
+        parentSpanId = value;
+      }
     } else {
       customContext = Object.assign({}, customContext, {
         [name]: value,
--- a/lib/propagation/aws.js
+++ b/lib/propagation/aws.js
@@ -15,7 +15,7 @@ function marshalTraceContextv1(context) {
   let spanId = context.parentSpanId || util.currentSpanId(context);
   let traceFields = context.customContext || context.traceContext || {};
 
-  return `Root=${traceId};Parent=${spanId}${util.objToString(traceFields, ";")}`;
+  return `Root=${traceId};Parent=${spanId};${util.objToString(traceFields, ";")}`;
 }
 
 // unmarshalTraceContext takes a string trace header and returns a context
--- a/lib/propagation/aws.js
+++ b/lib/propagation/aws.js
@@ -79,8 +79,6 @@ function unmarshalTraceContext(header) {
 
 exports.httpTraceParserHook = httpTraceParserHook;
 
-// parserHook functions do not require users to pass an input
-// the instrumentation will provide the request object
 function httpTraceParserHook(httpReq) {
   const header = httpReq.headers["x-amzn-trace-id"];
   if (!header) {
--- a/lib/propagation/aws.js
+++ b/lib/propagation/aws.js
@@ -18,7 +18,7 @@ function marshalTraceContextv1(context) {
   let contextToSend;
 
   if (Object.keys(traceFields).length > 0) {
-    contextToSend = `;${util.objToString(traceFields, ";")}`;
+    contextToSend = util.objToString(traceFields, ";", "=", ";");
   }
 
   return `Root=${traceId};Parent=${spanId}${contextToSend}`;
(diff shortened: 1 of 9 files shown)


lib/propagation/util.js
```diff
@@ -23,3 +23,39 @@ function getPropagationContext(context = {}) {
...
 }
+
+exports.objToString = objToString;
+
+// pass this util an object, return as string formatted to go into a trace header
+// joinChar separates fields from each other, keyChar separates keys from values
+// function is written to place a joinChar at the beginning of the string
+function objToString(obj, joinChar = ",", keyChar = "=") {
+  let string = "";
+  const keys = Object.keys(obj) || {};
```
katiebayes, line 34: Just peeking at this, realized this should fallback to an empty array, not an empty Object.
```suggestion
  const keys = Object.keys(obj) || [];
```

lib/propagation/util.js
```diff
@@ -23,3 +23,39 @@ function getPropagationContext(context = {}) {
     customContext: context.traceContext,
   };
 }
+
+exports.objToString = objToString;
+
+// pass this util an object, return as string formatted to go into a trace header
+// joinChar separates fields from each other, keyChar separates keys from values
+// function is written to place a joinChar at the beginning of the string
```
katiebayes, line 31: No longer true! Now uses `join()` and will return a string without a leading char. This was done for flexibility on ordering. The needed character is added in the string literal in the marshal function instead.
```suggestion
```

lib/propagation/aws.js
```diff
@@ -60,3 +70,19 @@ function unmarshalTraceContext(header) {
     customContext,
   };
 }
+
+exports.httpTraceParserHook = httpTraceParserHook;
+
+// parserHook functions do not require users to pass an input
+// the instrumentation will provide the request object
+function httpTraceParserHook(req = {}) {
```
toshok, line 78: the default parameters on these hooks bug me.  I don't think they help ultimately guard against exceptions being thrown, and instead only change where that exception _is_ thrown.

if req is undefined we'll end up with {}, and `header` below will be undefined.  we'll pass that `undefined` to  `unmarshalTraceContext`, which will throw an exception because the first thing it does is call `header.split`.

why are we trying to support objects that _don't_ have a `headers` field?

barring another reason to have the {} fallback, imo the parser hooks should look like:

```
function httpTraceParserHook(httpReq) {
  return unmarshalTraceContext(httpReq.headers["x-amzn-trace-id"]);
}
```

lib/propagation/aws.js
```diff
@@ -60,3 +70,19 @@ function unmarshalTraceContext(header) {
...
+  const header = req.headers && req.headers["x-amzn-trace-id"];
+
+  return unmarshalTraceContext(header);
+}
+
+exports.httpTracePropagationHook = httpTracePropagationHook;
+
+function httpTracePropagationHook(context = {}) {
+  return { [TRACE_HTTP_HEADER]: exports.marshalTraceContext(context) };
```
toshok, line 87:
```suggestion
  return { [TRACE_HTTP_HEADER]: marshalTraceContextv1(context) };
```

lib/propagation/aws.js
```diff
@@ -60,3 +70,19 @@ function unmarshalTraceContext(header) {
...
+// the instrumentation will provide the request object
+function httpTraceParserHook(req = {}) {
+  const header = req.headers && req.headers["x-amzn-trace-id"];
+
+  return unmarshalTraceContext(header);
+}
+
+exports.httpTracePropagationHook = httpTracePropagationHook;
+
+function httpTracePropagationHook(context = {}) {
```
toshok, line 86: same here - why do we have a default?  it'll cause a nonsensical string to be returned, something like:

```
Root=undefined;Parent=undefined
```

lib/propagation/aws.js
```diff
@@ -13,34 +13,39 @@ function marshalTraceContextv1(context) {
...
 
-  if (context.traceContext) {
-    const elements = Object.keys(context.traceContext).map(key => {
-      return `${key}=${context.traceContext[key]}`;
-    });
-    contextToSend = ";" + elements.join(";");
+  let contextToSend;
+
+  if (Object.keys(traceFields).length > 0) {
+    contextToSend = `;${util.objToString(traceFields, ";")}`;
```
toshok, line 21: honestly I'd prefer to have `objToString` take another parameter (`prefix`) that this code passes (the default would be `""` since the other uses don't need it).  That would let us always call `objToString` here, and avoid the duplicate `Object.keys` call here (`objToString` already does it and returns "" if there are none.)

but more meh.

lib/propagation/util.js
```diff
@@ -23,3 +23,38 @@ function getPropagationContext(context = {}) {
...
+
+exports.objToString = objToString;
+
+// pass this util an object, return as string formatted to go into a trace header
+// joinChar separates fields from each other, keyChar separates keys from values
+function objToString(obj, joinChar = ",", keyChar = "=") {
+  let string = "";
+  const keys = Object.keys(obj) || [];
+
+  if (keys.length > 0) {
```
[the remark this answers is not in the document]
toshok: Actually, hust remove the keys.length > 0 check and always do the map+join.

toshok (changes requested):
a couple of comments with "meh" in them that I care less about.  They'd be clearer imo but I don't really need them.

lib/propagation/aws.js
toshok: actually even _that_ will throw an exception in `unmarshalTraceContext` if the header doesn't exist.

second try:

```
function httpTraceParserHook(httpReq) {
  const header = httpReq.headers["x-amzn-trace-id"];
  if (!header) {
    return null; // <- I'm also pretty down on naked returns in functions that also return values.  It's unclear if the intent was to return undefined or if the return value was forgotten.  in functions that return values, all returns should have them.
  }
  return unmarshalTraceContext(header);
}
```

paulosman:
@toshok took a crack at the "meh" comments (which I agreed with, so went ahead and changed). I'll work on the defaults next.

lib/propagation/aws.js
paulosman: Yeah, I agree with this - implemented.

lib/propagation/aws.js
paulosman: Also made `contextToSend` default to an empty string so we don't get `undefined` at the end of the header. I hope I did that right?

toshok (approved):
couple of remaining derp moments by yours truly :)

lib/propagation/util.js
```diff
@@ -23,3 +23,32 @@ function getPropagationContext(context = {}) {
...
+exports.objToString = objToString;
+
+// pass this util an object, return as string formatted to go into a trace header
+// joinChar separates fields from each other, keyChar separates keys from values
+function objToString(obj, joinChar = ",", keyChar = "=", prefix = "") {
+  const keys = Object.keys(obj);
+  const fields = keys.map(key => {
+    return key + keyChar + obj[key];
+  });
+  return prefix + fields.join(joinChar);
```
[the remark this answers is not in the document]
paulosman: Derp! Of course. I'll add this guard:

```
if (keys.length == 0) {
  return "";
}
```

PR #258, opened 2020-10-21
add a default for hook.js configure opt's parameter, and some bare bones hooks tests

toshok:
hooks.js's `configure` shouldn't throw an exception if passed undefined - all other configure methods default to `opts = {}`.  Add a test for the behavior of `configure()` and `configure({})`.

Also add some very contrived and basic tests for parser/propagation hooks (the hooks do no real work at all).  Leave as a comment some cases that are currently failing.  If we're going to actually make an assertion about the _type_ of the return value from the propagation hook, we've got a lot of possible holes.  Same holes exist in the type checking for the return value from `httpTraceParseHook`.

### DIFF
--- a/lib/propagation/hooks.js
+++ b/lib/propagation/hooks.js
@@ -8,7 +8,7 @@ let httpTracePropagationHook;
 
 exports.configure = configure;
 
-function configure(opts) {
+function configure(opts = {}) {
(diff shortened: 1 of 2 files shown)


paulosman (approved):
Thank you!

PR #259, opened 2020-10-22
update (several) deps

toshok:
grouped together several deps in commits that required no changes on our part.  for dependencies that required code changes I made them separate commits so you can browse the changes.  they were mostly minor (the largest required change was the `supertest` bump, which required some pretty invasive reworking of express/fastify tests, but the changes make sense.)

The remaining outdated direct deps are:

* `fastify`  it's a major bump, so this should be its own PR I expect.  Not a clue about API changes.
* `husky`/`lint-staged` these are annoying and husky in particular changed its behavior in the _three_ major versions we're behind.  single PR for these.
* `pg`/`pg-query-stream` dunno - but their own PR certain.
* `prettier` this one I'll do next.  going to do its own PR because it's going to whack a lot of files and didn't want to clutter up code changes with formatting changes.

### DIFF
--- a/lib/instrumentation/express.js
+++ b/lib/instrumentation/express.js
@@ -1,7 +1,7 @@
 /* eslint-env node */
 const onHeaders = require("on-headers"),
   shimmer = require("shimmer"),
-  flatten = require("array-flatten"),
+  { flatten } = require("array-flatten"),
   tracker = require("../async_tracker"),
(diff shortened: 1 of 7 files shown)


toshok:
weird.  `npm audit fix` doesn't seem to actually fix the audit?

paulosman (approved):

PR #262, opened 2020-10-23
return result of spanFn even when there is no parent context

SomeKittens:
There's a case where I want to asynchronously trace something but also care about the return value:

```javascript
const makeQuery = async (query) => {
  return beeline.startAsyncSpan({
    name: 'example',
  }, (span) => {
    return makeDatabaseQuery(query)
    .finally(() => beeline.finishSpan(span));
  });
};
```

This works just fine _so long as there's an active trace._  If there's no active trace, `startAsyncSpan` returns `undefined` and the whole thing falls over.  This PR ensures that the `startAsyncSpan` always returns the value returned from `spanFn`.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -6,3 +6,4 @@ Erwin van der Koogh
 Ally Weir
 Ashish Bista
 Mordy Tikotzky
+Randall Koutnik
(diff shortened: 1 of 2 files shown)


toshok (approved):
gah, nice find.  \o/

PR #263, opened 2020-10-28
Support fastify v3

martin308:
Update to v3 of Fastify, the current major version.

Still supports v2 of fustily but we don't have a good way of maintaining tests for multiple versions of dependencies 😢

### DIFF
--- a/lib/instrumentation/fastify.js
+++ b/lib/instrumentation/fastify.js
@@ -94,7 +94,7 @@ const instrumentFastify = function(fastify, opts = {}) {
         [schema.TRACE_SPAN_NAME]: "request",
         [schema.TRACE_ID_SOURCE]: spanContext.source,
         "request.host": request.hostname,
-        "request.original_url": request.req.originalUrl,
+        "request.original_url": request.raw.url,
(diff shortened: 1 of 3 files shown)


lib/instrumentation/fastify.js
```diff
@@ -94,7 +94,7 @@ const instrumentFastify = function(fastify, opts = {}) {
         [schema.TRACE_SPAN_NAME]: "request",
         [schema.TRACE_ID_SOURCE]: spanContext.source,
         "request.host": request.hostname,
-        "request.original_url": request.req.originalUrl,
+        "request.original_url": request.raw.url,
```
toshok, line 97: do we need to fallback here for older versions too?

```suggestion
        "request.original_url": request.raw ? request.raw.url : request.req.originalUrl,
```

maybe?

toshok (approved):

lib/instrumentation/fastify.js
martin308: It worked when I ran the tests with fastify v2 and with v3 in this form, but it could potentially be safer to have this  additional fallback

PR #272, opened 2020-12-23
Handle pg pool clients

markgaylard:
We found an issue when calling the promise returning version of `query` on a pg-pool client. These clients seem to behave a bit differently to both the query method on both `pg.Pool` and `pg.Client`. Both of those create a callback function before the the instrumentation is invoked, but in our case, the callback is created _after_ instrumentation. The upshot is that no callback is passed to the query function instrumented by beeline, so even though the spans are started, they are never finished.

This PR adds a wrapper to promises returned by the query function to fix this issue. It also expands on the tests for the pg instrumentation by asserting that the results returned by the wrapped functions are correct. I've also added a bit to the README file to explain how to run the pg tests.

### DIFF
--- a/lib/instrumentation/pg.js
+++ b/lib/instrumentation/pg.js
@@ -124,7 +124,7 @@ let instrumentPg = function (pg, opts = {}) {
           }
 
           const r = query.apply(this, [config, values, callback]);
-          return r instanceof Promise && !callback ? promiseWrapper(span, r) : r;
+          return r instanceof Promise ? promiseWrapper(span, r) : r;
         }
       );
     };
--- a/lib/instrumentation/pg.js
+++ b/lib/instrumentation/pg.js
@@ -46,7 +46,7 @@ const promiseWrapper = (span, promise) => {
           "db.error_hint": error.hint,
         });
       }
-      return error;
+      throw error;
     })
     .finally(() => api.finishSpan(span, "query"));
 };
--- a/lib/instrumentation/pg.js
+++ b/lib/instrumentation/pg.js
@@ -30,6 +30,27 @@ const getQueryName = function ([config]) {
   return undefined;
 };
 
+const promiseWrapper = (span, promise) => {
+  return promise
+    .then((result) => {
+      span.addContext({
+        "db.rows_affected": result.rowCount,
+      });
+      return result;
+    })
+    .catch((error) => {
+      if (error !== null && error instanceof Error) {
+        span.addContext({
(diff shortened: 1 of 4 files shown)


markgaylard:
Not sure why circle ci is sad. Doesn't look like it anything from this PR though.

MikeGoldsmith:
Thanks for the contribution @markgaylard - we'll review and get back to you ASAP.

I'm not sure why CI failed, it looks to be from an unrelated test that timed out. I've triggered the CI to run again.

lib/instrumentation/pg.js
```diff
@@ -30,6 +30,27 @@ const getQueryName = function ([config]) {
...
+    })
+    .catch((error) => {
+      if (error !== null && error instanceof Error) {
+        span.addContext({
+          "db.error": error.message,
+          "db.error_stack": error.stack,
+          "db.error_hint": error.hint,
+        });
+      }
+      return error;
```
vreynolds, line 49: `error` needs to be re-thrown here, otherwise the downstream promise chain will receive it in a `then` rather than a `catch`

lib/instrumentation/pg.js
```diff
@@ -102,7 +123,8 @@ let instrumentPg = function (pg, opts = {}) {
             }
           }
 
-          return query.apply(this, [config, values, callback]);
+          const r = query.apply(this, [config, values, callback]);
+          return r instanceof Promise && !callback ? promiseWrapper(span, r) : r;
```
vreynolds, line 127: Curious about the `&& !callback` here: I _think_ it might be unnecessary, since `pg` will only return a Promise, [when no callback has been defined](https://github.com/brianc/node-postgres/blob/3f3f1a77c3a87e42df64c5baaa7d42193b0d8529/packages/pg/lib/client.js#L515). Thoughts?

vreynolds (changes requested):
<3 the test cases, super helpful! Mostly LGTM, just a couple small asks.

lib/instrumentation/pg.js
markgaylard: Yeah, I can't remember why I added this. I think it might have been while i was testing stuff but it doesn't need to be there so I've removed it.

lib/instrumentation/pg.js
markgaylard: nice catch (see what I did there? 😄 ) 
I've fixed this and updated the tests.

vreynolds (approved):

PR #254, opened 2020-09-29
Add interop docs to API.md

paulosman:
- Adding the interop API functions to the API documentation
- Some general layout adjustments and adding a list of contents to ease findability

PR #286, opened 2021-01-13
add trace hooks API docs

vreynolds:
Follow-up to #254, adding API docs about trace parser/propagation hooks


paulosman (approved):
:evergreen_tree:

PR #287, opened 2021-01-14
Ensure fastify instrumentation honors the httpParserHook config

vreynolds:
* add test for this case to both fastify and express instrumentations

Fastify instrumentation was using `traceUtils.getTraceContext` which does not use `httpTraceParserHook` configuration.

### DIFF
--- a/lib/instrumentation/fastify.js
+++ b/lib/instrumentation/fastify.js
@@ -72,7 +72,7 @@ const instrumentFastify = function (fastify, opts = {}) {
       // start our trace handling
 
       // parse incoming trace headers into a span context
-      let spanContext = traceUtil.getTraceContext(traceIdSource, request);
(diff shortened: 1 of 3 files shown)


lib/instrumentation/fastify.js
```diff
@@ -72,7 +72,7 @@ const instrumentFastify = function (fastify, opts = {}) {
       // start our trace handling
 
       // parse incoming trace headers into a span context
-      let spanContext = traceUtil.getTraceContext(traceIdSource, request);
+      let spanContext = traceUtil.parseTraceHeader(traceIdSource, request);
```
robbkidd, line 75: Yay! ✨  consistency ✨

robbkidd (approved):
Getting the hook right!

Captain Hook, how's left hand is a hook, playing the piano.

Issue #271, opened 2020-12-21
FastifyDeprecation: request.req -> request.raw


LewisJEllis:
Getting this with Fastify:
```
(node:46560) [FSTDEP001] FastifyDeprecation: You are accessing the Node.js core request object via "request.req", Use "request.raw" instead.
```

Looks like it should just be a small `s/request.req/request.raw/g` tweak here: https://github.com/honeycombio/beeline-nodejs/blob/main/lib/instrumentation/fastify.js#L99-L103

vreynolds:
Fixed in [v2.6.0](https://github.com/honeycombio/beeline-nodejs/releases/tag/v2.6.0)

PR #288, opened 2021-01-15
Use the non-deprecated request.raw attribute

vreynolds:
* looks like request.req has been aliased to raw for a while: https://github.com/fastify/fastify/commit/4cb0a4b20a0d9bc6a6ffa6901b6b0d3c09f41a29
* this should continue to work for older versions of fastify, and no longer show a deprecation warning
* verified that fields are populated in Honeycomb with latest 3.x and a 2.0.0 fastify

Addresses #271


robbkidd (approved):
Quiet those deprecation warnings. 👍 

a person peacefully settling into a bathrub

PR #296, opened 2021-01-22
Prepare v2.6.0 release

MikeGoldsmith:
Updates package version to `2.6.0` and the change log with commits since last release.


robbkidd (approved):
a whole bunch of guinea pigs are released from an enclosure

Issue #126, opened 2019-03-08
Define Typescript Types


allyjweir:
When working with `beeline-nodejs`, understanding which functions are available on the `beeline` object imported can require a lot of referring back to the library's implementation.

Having done a fair bit of spelunking into the code, I'd be happy to take a pass at writing the initial types. Would this be something that the codebase would be happy to accept?

toshok:
absolutely 👍

marcusradell:
@allyjweir do you think this will happen or are you blocked by something? (For example time available for free coding.)

allyjweir:
Hey @marcusradell, we had got somewhere with it but never ready for production. Then I changed jobs so no longer working in Typescript so kind of dropped this issue. If you'd like to pick it up then by all means please do!

sampsakuronen:
Our team would like to request this as well! It's beneficial for Honeycomb API exploration (looking through the documentation vs. seeing directly in the IDE what are the possible configuration options).

alferpal:
Is anyone working on this?

Is it something of interest to the library?

I'm willing to prepare a types file from scratch ( or if there is anything already started, continue with it) on my free time (no ETA)

antonvasin:
I've had most of the definitions written some time ago but never had a chance to properly test them. I'll try to open PR soon

bkrausz:
Thanks for getting this out @antonvasin!

@vreynolds would we be able to get a new version release so we can start using this?

PR #308, opened 2021-02-09
TypeScript definitions

antonvasin:
As mentioned in #126.

When writing I've used [API.md](https://github.com/honeycombio/beeline-nodejs/blob/main/docs/API.md) as a reference.

Things I'm not sure about:
* `MetadataContext` seems to deserve stricter type than `any`
* Can context be nested? If so, then we should have some recursive type for it
* `type` and `interface` syntax. Some use both, some use only one, so those things can be very opinionated. Since the project itself is written in plain JS I've tried to write it “default” way
* Couldn't test this with `require` since we're not using it in TS projects at all


lib/types.d.ts
```diff
@@ -0,0 +1,146 @@
+import http from "http";
+
+export interface BeelineOpts {
```
vreynolds, line 3: Additional optional configurations:
- samplerHook
- presendHook
- disableInstrumentation
- express:
  - userContext
  - traceIdSource
  - parentIdSource
- fastify
  - userContext
  - traceIdSource
  - parentIdSource
- mongodb
  - includeDocuments

lib/types.d.ts
```diff
@@ -0,0 +1,146 @@
+import http from "http";
```
vreynolds, line 1: I'm getting a linter error here: `Module '"http"' has no default export.` Perhaps `import * as http from "http"` or just `import {IncomingMessage} from "http"`?

lib/types.d.ts
```diff
@@ -0,0 +1,146 @@
...
+  startTimeHR: [number, number];
+}
+
+export type MetadataContext = Record<string, any>;
+
+export interface TraceContext {
+  traceId?: string;
+  parentSpanId?: string;
+  dataset?: string;
+  customContext?: TraceContext;
```
vreynolds, line 45: I believe `customContext` here would be a `MetadataContext`. Is that the nested context question you had?

lib/types.d.ts
```diff
@@ -0,0 +1,146 @@
...
+  name: string;
+  startTimeHR: [number, number];
+}
+
+type SpanFn<F> = (...args: any[]) => F;
+
+type Configure = (opts?: BeelineOpts) => Beeline & Configure;
+
+type Headers = Record<string, string>;
+type HttpTraceParserHook = (req: http.IncomingMessage) => TraceContext;
```
vreynolds, line 58: There's a use case for creating custom trace parser/propagator hooks. Can we export `HttpTraceParserHook` and `HttpTracePropagatorHook` types?

lib/types.d.ts
```diff
@@ -0,0 +1,146 @@
...
+type HttpTracePropagationHook = (ctx: TraceContext) => Headers;
+
+export interface Beeline {
+  traceActive(): boolean;
+  clearTrace(): void;
+  getTraceContext(): TraceContext;
```
vreynolds, line 64: After digging through the code, I noticed that there's 2 "context" representations. There's the `TraceContext` as well another one (ExecutionContext??), used in the [async tracker](https://github.com/honeycombio/beeline-nodejs/blob/b6cf0963cbbc7effdaff77b40d8772ff5ed2e18a/lib/api/libhoney.js#L110), which is what's returned from `getTraceContext()`:
```
{
  id?: string;
  traceContext?: MetadataContext;
  stack?: Span[];
  dataset?: string;
}
```

This exercise is certainly highlighting some loose ends in the interface!

lib/types.d.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  startTrace(
+    metadataContext?: MetadataContext,
+    traceId?: string,
+    parentSpanId?: string,
+    dataset?: string
```
vreynolds, line 70: `startTrace` takes `propagatedContext?: MetadataContext` as a final argument, and also it looks like return type is `Span | undefined` based on this [line](https://github.com/honeycombio/beeline-nodejs/blob/b6cf0963cbbc7effdaff77b40d8772ff5ed2e18a/lib/api/libhoney.js#L131)

lib/types.d.ts
```diff
@@ -0,0 +1,146 @@
...
+  finishTrace(trace: Span): void;
+  withTrace<F>(
+    metadataContext: MetadataContext,
+    fn: SpanFn<F>,
+    traceId?: string,
+    parentSpanId?: string,
+    dataset?: string
+  ): F;
+
+  startSpan(metadataContext?: MetadataContext): Span;
```
vreynolds, line 81: `startSpan` returns `Span | undefined` based on this [line](https://github.com/honeycombio/beeline-nodejs/blob/b6cf0963cbbc7effdaff77b40d8772ff5ed2e18a/lib/api/libhoney.js#L131)

lib/types.d.ts
```diff
@@ -0,0 +1,146 @@
...
+  finishSpan(event: Span): void;
```
vreynolds, line 82: `finishSpan` takes `rollup?: string` as a second argument

lib/types.d.ts
```diff
@@ -0,0 +1,146 @@
...
+  withSpan<F>(metadataContext: MetadataContext, fn: SpanFn<F>, rollup?: string): F;
+  withSpan<F>(fn: SpanFn<F>, rollup?: string): F;
```
vreynolds, line 84: Is the second `withSpan` line intentional? I think metadataContext is non-optional: https://github.com/honeycombio/beeline-nodejs/blob/b6cf0963cbbc7effdaff77b40d8772ff5ed2e18a/lib/api/index.js#L118

lib/types.d.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  bindFunctionToTrace<F>(fn: SpanFn<F>): F;
+  runWithoutTrace<F>(fn: SpanFn<F>): F;
+
+  flush(): Promise<void>;
+
+  getInstrumentations(): string[];
+
+  /** @deprecated this method will be removed in the next major release. Please use honeycomb.marshalTraceContext() instead. */
+  marshalTraceContext(ctx: TraceContext): string;
```
vreynolds, line 111: Based on above comment about the 2 context representations, all the marshal functions are written to work with both, so I think they would take `TraceContext | ExecutionContext` as an argument.

lib/types.d.ts
```diff
@@ -0,0 +1,146 @@
...
+  /** @deprecated this method will be removed in the next major release. Please use honeycomb.unmarshalTraceContext() instead. */
+  unmarshalTraceContext(ctx: string): TraceContext;
```
vreynolds, line 113: All the unmarshal functions may return `undefined` in case of an error.

lib/types.d.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  honeycomb: {
+    marshalTraceContext(ctx: TraceContext): string;
+    unmarshalTraceContext(ctx: string): TraceContext;
```
vreynolds, line 117: Tiny pick: `unmarshalTraceContext` argument is a `honeycombTraceHeader` not a `ctx`

lib/types.d.ts
```diff
@@ -0,0 +1,146 @@
...
+    httpTraceParserHook: HttpTraceParserHook;
+    httpTracePropagationHook: HttpTracePropagationHook;
+    TRACE_HTTP_HEADER: string;
+  };
+
+  w3c: {
+    marshalTraceContext(ctx: TraceContext): string;
+    unmarshalTraceContext(traceparent: string, tracestate: string): TraceContext;
```
vreynolds, line 125: `tracestate` can be optional

lib/types.d.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  aws: {
+    marshalTraceContext(ctx: TraceContext): string;
+    unmarshalTraceContext(amazonTraceHeader: string): TraceContext;
+    httpTraceParserHook: HttpTraceParserHook;
+    httpTracePropagationHook: HttpTracePropagationHook;
+    TRACE_HTTP_HEADER: string;
+  };
+
+  TRACE_HTTP_HEADER: string;
```
vreynolds, line 139: `TRACE_HTTP_HEADER` is deprecated: `/** @deprecated this constant will be removed in the next major release. Please use honeycomb.TRACE_HTTP_HEADER instead. */`

lib/types.d.ts
```diff
@@ -0,0 +1,146 @@
...
+  httpTraceParserHook: HttpTraceParserHook;
```
vreynolds, line 140: I don't believe there's a top-level `httpTraceParserHook`/`httpTracePropagationHook` exported. I think you already defined all of them under the 2 namespaces (honeycomb, w3c, aws)

vreynolds (changes requested):
Thank you so much @antonvasin, this will definitely be appreciated!

I did my best to capture any discrepancies, combing through the code. As you can see, the docs are not super comprehensive, unfortunately.

Do you have bandwidth to address the changes?

vreynolds:
Oh, as far as your questions:
- MetadataContext: I agree, it would be nice to nail down exactly the attribute types, but I think at this point, I'd rather err on the side of permissive/backwards-compatible way (aka `any`)
- nested context: I think I addressed that in the review comment, but let me know if you meant something else
- type and interface: LGTM
- I used `require` to import the beeline in my test TS service

antonvasin:
Thanks for the feedback.

> Do you have bandwidth to address the changes?

Sure, I'll look into them.

lib/types.d.ts
```diff
@@ -0,0 +1,146 @@
...
+export interface Span {
+  addContext(metadataContext: MetadataContext): void;
+  payload: Schema & MetadataContext;
+  startTime: number;
+  startTimeHR: [number, number];
+}
+
+export type MetadataContext = Record<string, any>;
+
+export interface TraceContext {
```
joerayme, line 41: I believe this can also have an optional `source` attribute, [for example](https://github.com/honeycombio/beeline-nodejs/blob/28a176981e38a23d034d4a9e7a7f99b15daa18c8/lib/instrumentation/trace-util.js#L26)

DavidS:
see https://github.com/honeycombio/libhoney-js/pull/90 for a similar effort.

lib/types.d.ts
antonvasin: Yeah, my bad. Removed those definitions.

lib/types.d.ts
antonvasin: Added `@deprecated` to this

lib/types.d.ts
antonvasin: Renamed this

lib/types.d.ts
antonvasin: I've defined `MarshallableContext` type for those and updated argument type.

antonvasin:
I've updated this MR with comments above. Took me a while, sorry.

vreynolds (approved):

vreynolds:
@antonvasin same here, took a while to come back to this. The changes look good to me 👍 Do you mind resolving the package conflict?

PR #339, opened 2021-04-16
transmission is a property of libhoney, but if not declared we cannot pass it from typescript

ilbambino:


vreynolds (approved):

ilbambino:
so fast, thank you

PR #348, opened 2021-05-07
Make api.withTrace's type signature match api.startTrace

ajvondrak:
Previously, `api.withTrace` did not accept a `propagatedContext`
parameter, so the call to `apiImpl.startTrace` was pushing a context
stack record with an `undefined` trace context. This would break
`api.addTraceContext` with the error:

    honeycomb-beeline/lib/api/libhoney.js:314
      Object.assign(context.traceContext, map);
             ^
    TypeError: Cannot convert undefined or null to object

This patch fixes the issue by making the type signature of
`api.withTrace` match that of `api.startTrace`, thereby giving the
`propagatedContext` parameter a default value that gets passed along.

`api.addTraceContext` isn't exercised very thoroughly by the tests
as-is, but I added a quick regression test for the synchronous API bug.

### DIFF
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -100,8 +100,21 @@ module.exports = {
   finishTrace(trace) {
     return apiImpl.finishTrace(trace);
   },
-  withTrace(metadataContext, fn, withTraceId, withParentSpanId, withDataset) {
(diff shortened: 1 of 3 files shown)


vreynolds:
LGTM! @ajvondrak can you rebase?

ajvondrak:
@vreynolds Pushed!

vreynolds (approved):

PR #351, opened 2021-05-11
docs: show how to import beeline in TypeScript

jazzdan:
If you use `require` you don't get any of the TypeScript support

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -16,8 +16,7 @@ const beeline = require("honeycomb-beeline")();
 Or in TypeScript:
 
 ```typescript
-import honeycomb from "honeycomb-beeline";
-const beeline = honeycomb();
+import beeline = require('honeycomb-beeline');
 
 /* beeline.traceActive() works, as do all the other calls below */
 ```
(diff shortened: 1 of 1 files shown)


vreynolds:
I'll take a look at the types definition. I hesitate to add just the `import` bit, because the beeline initializer needs to be `require`d. So this is fine, as long as TS compiles to commonjs module.

vreynolds:
https://github.com/honeycombio/beeline-nodejs/pull/352 should make TS happier with either require or import, but there is a caveat that with require, it needs special syntax:
```js
import beeline = require('honeycomb-beeline');
```

@jazzdan does it make sense to update this PR to reflect that? and we will merge it after the above fix goes through the pipes.

jazzdan:
@vreynolds yeah for sure, I can do that.

Is there a way to import the beeline as it currently exists such that TypeScript is aware of the types _and_ the beeline gets initialized?

vreynolds:
@jazzdan I think what you suggested originally should work, as long as your tsconfig compiles to module: commonjs.

vreynolds (approved):

jazzdan:
ah, gotcha, Thanks @vreynolds!

PR #352, opened 2021-05-11
Ensure beeline can be import-ed or require-d when using typescript

vreynolds:
* note: require will look like `import beeline = require('honeycomb-beeline')`
* https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-function-d-ts.html

The diff looks awful, this is just:
- add a namespace around the whole thing
- update default export

### DIFF
--- a/lib/types.d.ts
+++ b/lib/types.d.ts
@@ -1,181 +1,184 @@
 import { IncomingMessage } from "http";
 
-export interface SamplerResponse {
-  sampleRate: number;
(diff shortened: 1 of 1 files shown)


robbkidd:
Highly recommend [reviewing the diff with whitespace ignored](https://github.com/honeycombio/beeline-nodejs/pull/352/files?diff=unified&w=1)!

robbkidd (approved):
bee love

PR #360, opened 2021-05-24
Add node 16 to test matrix, switch to non-legacy circle images

vreynolds:
See https://github.com/honeycombio/libhoney-js/pull/135 for reasons to switch to `cimg` and dropping older odd versions.


paulosman (approved):
:dancers:

PR #361, opened 2021-05-26
Fix: http(s) instrumentation URL arg

vreynolds:
* alternative to #324 
* http requests can be constructed with a WHATWG URL type object
* add handling for URLs and test argument variations


paulosman (approved):
LGTM!

PR #401, opened 2021-07-14
Use JSON.stringify only when debugging is enabled

kamilkisiela:
There's no need to call `JSON.stringify` for every finished span when debugging is disabled. It can be done lazily.

If `JSON.stringify` takes 0.1ms on average then for 50 spans we get 5ms added to the overall latency.


MikeGoldsmith (approved):
This change looks good - thanks for the contribution @kamilkisiela 👍🏻

Issue #172, opened 2019-11-06
Webpack/bundler support?


nason:
I'm trying to get some of our lambda services instrumented. They're written in node and bundled with webpack (via https://github.com/serverless-heaven/serverless-webpack)

Is this a supported scenario? I'm running into issues (such as https://github.com/honeycombio/beeline-nodejs/issues/128) that give me the impression its not.

toshok:
yeah, bundlers won't work unless we can shim their require-logic in order to instrument packages.

charsleysa:
Can `pkg = require(path.join(__dirname, "..", "..", "package.json")),` be replaced with `pkg = require("../../package.json"),`?

chaoyangnz:
this library is very unfriendly to webpack bundle, like @charsleysa mentioned, it is using this kind of dynamic require and makes webpack unable to resolve at compile time.

please refine this library and comply the industry standard. I know this is a nodejs library, will you never audit how many dependencies this library is using and how is it used by your customers?

reconbot:
Going to add esbuild has similar problems

vreynolds:
Hello,

We will be closing this issue as it is a low priority for us. It is unlikely that we'll ever get to it, and so we'd like to set expectations accordingly.

As we enter 2022 Q1, we are trimming our OSS backlog. This is so that we can focus better on areas that are more aligned with the OpenTelemetry-focused direction of telemetry ingest for Honeycomb.

If this issue is important to you, please feel free to ping here and we can discuss/re-open.

PR #415, opened 2021-08-03
Support esbuild

markandrus:
After applying the changes in this PR, I was able to generate traces from a bundle using esbuild. I have not tested webpack, since I don't use it, but this should at least partially address #172.

My test app manually calls `startTrace`/`finishTrace` and uses Node's `http` library. You can see below that, once restructured, esbuild is able to handle the `require` calls for instrumentation:

```
$ esbuild --bundle lib/index.js --platform=node | grep 'instrumentations = {' -A 14
    var instrumentations = {
      bluebird: () => require_bluebird(),
      child_process: () => require_child_process(),
      express: () => require_express(),
      fastify: () => require_fastify(),
      http: () => require_http2(),
      https: () => require_https2(),
      mongodb: () => require_mongodb(),
      mongoose: () => require_mongoose(),
      mpromise: () => require_mpromise(),
      mysql2: () => require_mysql2(),
      pg: () => require_pg(),
      "react-dom/server": () => require_react_dom_server(),
      sequelize: () => require_sequelize()
    };
```

**EDIT:** In practice, this means that, although honeycomb-beeline can now be used from esbuild bundles, the only auto-instrumentations supported will be for built-in libraries (at least when used with `--platform=node`). See this comment for details: https://github.com/honeycombio/beeline-nodejs/pull/415#issuecomment-907116116

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,10 @@ This package makes it easy to instrument your Express/NodeJS application to send
 
 - **Node 10+**
 
+## Known Issues
+
+- Using a bundler (esbuild, webpack, etc.) with the Beeline is unsupported. You may be able to use the Beeline with a bundler, but auto-instrumentations will likely not work.
+
 ## Contributions
 
 Features, bug fixes and other changes to `beeline-nodejs` are gladly accepted. Please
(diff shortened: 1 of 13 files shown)


JamieDanielson:
Hi @markandrus,

Thank you for the contribution! We will take a look through the changes and let you know if any questions come up.

vreynolds (approved):
It's a step in the right direction. Added a disclaimer in the readme, since it's less obvious now that bundled apps are not fully supported.

PR #465, opened 2021-09-29
clean up npm tarball

vreynolds:

- https://blog.npmjs.org/post/165769683050/publishing-what-you-mean-to-publish.html
- NPM tarball has all kinds of unnecessary stuff in it (check out `explore` tab: https://www.npmjs.com/package/honeycomb-beeline)
- it's best to have an allow list of things we *want* in the tarball

- NPM always includes readme and license
- it used to include notice and changelog but now it does not

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,12 @@
   "engines": {
     "node": ">= 10.*"
   },
+  "files": [
+    "docs",
+    "lib",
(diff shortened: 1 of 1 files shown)


robbkidd (approved):
I'll put it on the list.

PR #494, opened 2021-11-09
Migrate and update example from examples repo

JamieDanielson:

- migrates node-tracing example from examples repo

- simplifies example (no need for third party APIs)
- updated to use local beeline code as dependency, which is installed and linked with `npm run setup`
- removed deprecated methods in favor of new, e.g. `beeline.honeycomb.marshalTraceContext`
- Add custom span and fixed TraceContext example
- set everything to run as `npm start`


vreynolds (approved):

JamieDanielson:
@vreynolds requesting a re-review as I tore out some of the original example and added some new stuff in.

examples/node-tracing/app.js
```diff
@@ -0,0 +1,82 @@
+/* eslint-env node */
+"use strict";
+const beeline = require("honeycomb-beeline");
+
+const HONEYCOMB_DATASET = `${process.env.HONEYCOMB_DATASET}` || "tracing-example";
```
vreynolds, line 5: string interpolation results in a literal string "undefined" when there's no env var

```suggestion
const HONEYCOMB_DATASET = process.env.HONEYCOMB_DATASET || "tracing-example";
```

examples/node-tracing/app.js
```diff
@@ -0,0 +1,82 @@
...
+const SERVICE_NAME = `${process.env.SERVICE_NAME}` || "wall";
```
vreynolds, line 6:
```suggestion
const SERVICE_NAME = process.env.SERVICE_NAME || "wall";
```

examples/node-tracing/app.js
```diff
@@ -0,0 +1,82 @@
...
+const HONEYCOMB_API_KEY = `${process.env.HONEYCOMB_API_KEY}` || "abc123";
```
vreynolds, line 7:
```suggestion
const HONEYCOMB_API_KEY = process.env.HONEYCOMB_API_KEY || "abc123";
```

vreynolds (changes requested):

vreynolds (approved):
✨🍾✨

PR #502, opened 2021-11-26
Expand type definitions to include all libhoney options

glenjamin:

- The typescript type defininition is currently missing some fields that the beeline will pass through to libhoney

This adds all parameters documented in libhoney to the typescript definition.

The specific one I wanted to use was `apiHost`, but I figured I'd add them all while I was here.

`userAgentAddition` is now special-cased, so that the beeline addition is still included if the option is supplied.


MikeGoldsmith (approved):
Looks great, thank you for the contribution @glenjamin 👍🏻

PR #525, opened 2022-01-04
fix: check parent exists during instrumentation:

Asgaroth:

`parent` can be null depending on how a module was loaded, see:
https://github.com/nodejs/modules/issues/469

Particularly in nextjs a config file is loaded that causes `parent` to be null and fail the app from building with:

```
> Build error occurred
TypeError: Cannot read property 'id' of undefined // if (parent.id.includes(`node_modules/${pkg.name}`) || !enabledInstrumentations.has(loadRequest)) {
    at exports.instrumentLoad (/path/node_modules/honeycomb-beeline/lib/instrumentation.js:74:14)
    at Function._load (/path/node_modules/honeycomb-beeline/lib/instrumentation.js:158:16)
    at ModuleWrap.<anonymous> (internal/modules/esm/translators.js:199:29)
    at ModuleJob.run (internal/modules/esm/module_job.js:183:25)
    at async Loader.import (internal/modules/esm/loader.js:178:24) {
  type: 'TypeError'
}
```

- Skip instrumentation if `parent` is not defined for a module

### DIFF
--- a/lib/instrumentation.js
+++ b/lib/instrumentation.js
@@ -71,7 +71,7 @@ function getPackageVersion(request, options) {
 }
 
 const instrumentLoad = (exports.instrumentLoad = (mod, loadRequest, parent, opts = {}) => {
-  if (parent.id.includes(`node_modules/${pkg.name}`) || !enabledInstrumentations.has(loadRequest)) {
+  if (!parent || parent.id.includes(`node_modules/${pkg.name}`) || !enabledInstrumentations.has(loadRequest)) {
     // no magic here
     return mod;
   }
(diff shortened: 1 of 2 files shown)


MikeGoldsmith (approved):
Looks good, thanks @Asgaroth. I added a test to ensure there isn't a future regression.

MikeGoldsmith:
Hi @Asgaroth - [beeline-nodejs v3.2.1](https://github.com/honeycombio/beeline-nodejs/releases/tag/v3.2.1) is now available with your fix 👍🏻

Asgaroth:
Thanks a lot!

PR #557, opened 2022-03-31
Make api.addContext work in impl:mock mode

glenjamin:

- the `addContext` API wasn't working when using `impl: mock`
- it appeared to still be adding fields to the span in the "old way"

- updated the `libhoney` backend to use `span.addContext` under the hood, and then synced the `mock` backend to match

There were no existing tests for the `mock` backend, so I haven't added any new ones.


MikeGoldsmith (approved):
Looks good, thanks @glenjamin 👍🏻

PR #566, opened 2022-04-08
maint: remove lint-staged

vreynolds:

- we removed lint-staged from libhoney because it caused more confusion than good in recent past

- remove lint-staged here as well


JamieDanielson (approved):
cat closing door bye

PR #576, opened 2022-04-26
Add meta.span_type to root/subroot spans

### LINKED ISSUE #575: Categorize span types on send

MikeGoldsmith:
Adds support for recording whether a span is a root or subroot span within the trace. Root means it is the very first span across all services and subroot means it is the first span within a given service.

When starting a new trace manually, via instrumentation or via propagation, `startTrace` is used so this will ensure that meta.span_type is correctly set.

- Closes #575

- Adds meta.span_type field to lib/schema
- Sets the meta.span_type field when calling startTrace, with either root or subroot depending on if parentSpanId is set
- Add tests to ensure the field is set correctly

### DIFF
--- a/lib/api/index.test.js
+++ b/lib/api/index.test.js
@@ -829,7 +829,6 @@ test("addTraceContext works inside of withTrace", () => {
 });
 
 test("startTrace without parent ID sets meta.span_type as root", () => {
-  const honey = api._apiForTesting().honey;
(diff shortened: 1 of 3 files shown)


robbkidd (approved):
Thank you, little root! Please stay strong!

JamieDanielson (approved):
Looks good!

PR #616, opened 2022-09-15
[docs] developing with a local version of libhoney in the beeline's example application

emilyashley:

- When developing changes for beeline and libhoney concurrently, I wanted to run both at the same time locally in my development environment (rather than relying on published versions of libhoney). This provides documentation on how I did that :) 

- reduce complexity of npm script in example app
- add documentation for local development

### DIFF
--- a/examples/node-tracing/package.json
+++ b/examples/node-tracing/package.json
@@ -5,7 +5,7 @@
   "main": "app.js",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "setup": "npm install && cd ../.. && npm install",
+    "setup": "npm install && (cd ../.. && npm install)",
     "start": "node app.js",
     "start-all": "node app.js & node analysis.js"
   },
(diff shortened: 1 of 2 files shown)


examples/node-tracing/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "app.js",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "setup": "npm install && npm install ../../ && npm link ../../ && npm link honeycomb-beeline",
+    "setup": "npm install && cd ../.. && npm install",
```
robbkidd, line 8:
```suggestion
    "setup": "npm install && (cd ../.. && npm install)",
```

robbkidd (approved):
👍🏻 Works on my machine, which in this case is the point.

Works for me!


I added the suggestion for changing directory in a subshell with `(cd ../.. && … )` below, but also I found that `npm run setup` had me still in the `examples/node-tracing/` directory when it completed. Including the parens doesn't appear to change the behavior, but would only prevent people like me from suggesting seems-like-a-good-idea subshell parens in the future. 😁

PR #726, opened 2024-03-01
feat: support classic ingest keys, upgrade to libhoney 4.2.0

cewkrupa:

We [recently](https://github.com/honeycombio/libhoney-js/releases/tag/v4.2.0) updated libhoney-js to support classic ingest keys. This PR updates the beeline to this version of libhoney-js and swaps `isClassic` to use the new exported method from libhoney.

- bumps libhoney-js version from `4.0.0` to `4.2.0`
- removes `utils.isClassic` in favor of `libhoney.isClassic`
- Adds two tests that use classic/non-classic ingest keys to verify it works

### DIFF
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -5,6 +5,7 @@ const process = require("process"),
   tracker = require("../async_tracker"),
   pkg = require("../../package.json"),
   debug = require("debug")(`${pkg.name}:event`),
+  libhoney = require("libhoney"),
   LibhoneyImpl = require("./libhoney"),
   MockImpl = require("./mock"),
   utils = require("../util");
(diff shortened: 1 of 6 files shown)


lib/api/index.js
```diff
@@ -5,9 +5,9 @@ const process = require("process"),
   tracker = require("../async_tracker"),
   pkg = require("../../package.json"),
   debug = require("debug")(`${pkg.name}:event`),
+  libhoney = require("libhoney"),
```
MikeGoldsmith, line 8: Do we need two require's for `libhoney` and `LibhoneyImpl`?

PS I prefer `libhoney` over `LibhoneyImpl` if we can just keep line.

MikeGoldsmith (approved):
Looks good. Left one query about not duplicating libhoney reqires, but otherwise good 👍🏻

lib/api/index.js
cewkrupa: Yes I think we do need both, since the `LibhoneyImpl` is coming from `./libhoney.js` in this repo while `libhoney` is coming direct from the libhoney-js package as `libhoney`.

MikeGoldsmith (approved):
Looks good -  thanks @cewkrupa

